PR #115, opened 2025-06-08
Add Prettier to keep things consistent

kraenhansen:
Merging this PR will:
- Introduce the `prettier` CLI and two package scripts for the root package, one to check and one to write (/ fix).

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -7,12 +7,13 @@ import tseslint from "typescript-eslint";
 import eslintConfigPrettier from "eslint-config-prettier/flat";
 
 export default tseslint.config(
-  globalIgnores([".nx/**"]),
-  globalIgnores(["**/dist/**"]),
-  globalIgnores(["apps/test-app/ios/**"]),
-  globalIgnores(["packages/host/hermes/**"]),
-  globalIgnores(["packages/node-addon-examples/examples/**"]),
-  globalIgnores(["packages/ferric-example/ferric_example.d.ts"]),
+  globalIgnores([
+    "**/dist/**",
+    "apps/test-app/ios/**",
+    "packages/host/hermes/**",
+    "packages/node-addon-examples/examples/**",
+    "packages/ferric-example/ferric_example.d.ts",
+  ]),
   eslint.configs.recommended,
   tseslint.configs.recommended,
   eslintConfigPrettier,


.prettierignore
shirakaba, line 1: This seems to do the trick. I've tried entering various gitignored folders with "format on save" enabled and it successfully ignores them. It seems `node_modules` is implicitly ignored as well. 👍 

I'm hoping to get permission to opensource a tool I made at work that flattens all `.gitignores` under a tree into a single one – I made it specifically to support Prettier, which has spent years neglecting monorepo support. But let's not wait on that.

.github/workflows/check.yml
```diff
@@ -18,6 +18,7 @@ jobs:
           node-version: lts/jod
       - run: npm ci
       - run: npm run lint
+      - run: npm run prettier:check
```
shirakaba, line 21:
```suggestion
      - run: node --run lint
      - run: node --run prettier:check
```

package.json
```diff
@@ -32,10 +34,12 @@
     "@tsconfig/react-native": "3.0.5",
     "@types/node": "^22.13.0",
     "eslint": "^9.19.0",
+    "eslint-config-prettier": "^10.1.5",
     "globals": "^16.0.0",
+    "prettier": "3.5.3",
```
shirakaba, line 39: Let's upgrade this to Prettier 3.6 (see [blog post](https://prettier.io/blog/2025/06/23/3.6.0)) for mad speed.

package.json
```diff
@@ -13,6 +13,8 @@
     "clean": "tsc --build --clean",
     "dev": "tsc --build --watch",
     "lint": "eslint .",
+    "prettier:check": "prettier --check .",
+    "prettier:write": "prettier --write .",
```
shirakaba, line 17: [From Prettier 3.6](https://prettier.io/blog/2025/06/23/3.6.0#cli), we can use the [performance-improved CLI](https://prettier.io/blog/2023/11/30/cli-deep-dive).

```suggestion
    "prettier:check": "prettier --experimental-cli --check .",
    "prettier:write": "prettier --experimental-cli --write .",
```

.prettierrc
```diff
@@ -0,0 +1 @@
+{}
```
shirakaba, line 1: I'd suggest renaming to `prettier.config.mjs` for some Intellisense and adding [@prettier/plugin-oxc](https://github.com/prettier/prettier/tree/main/packages/plugin-oxc) for mad speed, which was [announced](https://prettier.io/blog/2025/06/23/3.6.0#change-17472) alongside Prettier 3.6.

After running:

```sh
npm add -D prettier@latest @prettier/plugin-oxc
```

```suggestion
/**
 * @satisfies {import("prettier").Config}
 * @see https://prettier.io/docs/en/configuration.html
 */
const config = {
  plugins: ["@prettier/plugin-oxc"],
};

export default config;
```

shirakaba (approved):
Looks good, just some suggestions!

.github/workflows/check.yml
kraenhansen: I plan to do a separate PR to update `npm run` across the repo 👍 Especially when calling other scripts within scripts.

PR #137, opened 2025-06-22
Eliminate npm install noise

matthargett:
It looks like upstream RN has some intrinsic noise that hasn't been cleaned up in a while, but we can make our corner of the OSS univerise a little tidier.

Eliminate `npm audit` and depcrecated package WARN noise on first npm install.

Tested on Windows 11, `npm test` passes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "packages": {
     "": {
       "name": "@react-native-node-api/root",
+      "hasInstallScript": true,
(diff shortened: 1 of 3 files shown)


kraenhansen:
Obsolete now that I merged https://github.com/callstackincubator/react-native-node-api/pull/381 migrating the monorepo to pnpm (to get better primitives for running our scripts in the correct order and failing fast).

PR #143, opened 2025-06-25
Assert logs and run test suites conditionally

### LINKED ISSUE #141: Enable running suites in the test app conditionally
Our test app is already depending on a lot of native Node-API addons being built and this will only grow.

I see little need for building all addons if I only want to test a subset and I'd like to be able to run the test suites conditionally.

I'd like to leverage the Mocha Remote context API for this, allowing us to control which subset of the tests gets ran, from the commandline.

An additional stretch would be to expose selecting which suite to run based on GitHub labels on a PR.

kraenhansen:
Merging this PR will:
- Assert log output from Node Addon examples
- Run tests conditionally based on Mocha Remote context

This also fixes #141.


packages/node-addon-examples/src/index.ts
```diff
@@ -0,0 +1,101 @@
...
+    }
+  } finally {
+    console.log = originalLog;
+  }
+  // Throw and first error
+  const [firstError] = errors;
+  if (firstError) {
+    throw firstError;
+  }
+}
```
shirakaba, line 29: I like the lightweight assertions 👍

packages/node-addon-examples/src/index.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export const suites: Record<string, Record<string, () => void>> = {
```
shirakaba, line 31: I wasn't clear how/when this test file would be run, and whether it'd be in Node.js or React Native. May be worth adding some brief docs in `packages/node-addon-examples/README.md` and `apps/test-app/README.md`.

apps/test-app/App.tsx
```diff
@@ -8,22 +8,44 @@ import {
...
+type Context = {
+  allTests?: boolean;
+  nodeAddonExamples?: boolean;
+  ferricExample?: boolean;
+};
+
+function loadTests({
+  allTests = false,
+  nodeAddonExamples = allTests,
+  ferricExample = allTests,
```
shirakaba, line 30: So by default, both `nodeAddonExamples` and `ferricExample` resolve to `false` and so both the "Node Addon Examples" and "ferric-example" tests get skipped? Is that the intention?

shirakaba (approved):
Some minor comments!

apps/test-app/App.tsx
kraenhansen: I might still be debating this with myself 😊

On one hand it might be confusing to skip so many tests by default and on the other it does leave a potential to cut down on the time pre-building while bootstrapping and complexity (having to have the Rust toolchain installed) when someone wants to contribute to the project.

Looking back at this (I wrote it some weeks back), the optimization does seem a bit premature.

packages/node-addon-examples/src/index.ts
kraenhansen: I revisited the [packages/node-addon-examples/README.md](https://github.com/callstackincubator/react-native-node-api/blob/main/packages/node-addon-examples/README.md).

PR #168, opened 2025-07-10
Implementation of V1 buffer runtime functions

paradowstack:
This PR brings in an implementation of following `V1` buffer runtime functions:

- `napi_create_buffer`
- `napi_create_buffer_copy`
- `napi_is_buffer`
- `napi_get_buffer_info`
- `napi_create_external_buffer`

It also introduces test addon, adopted from https://github.com/nodejs/node/tree/main/test/node-api/test_buffer, that is placed in `packages/node-addon-examples/tests/buffers`.

### DIFF
--- a/packages/host/cpp/RuntimeNodeApi.cpp
+++ b/packages/host/cpp/RuntimeNodeApi.cpp
@@ -11,6 +11,10 @@ napi_status NAPI_CDECL callstack::nodeapihost::napi_create_buffer(
     return status;
   }
 
+  // Warning: The returned data structure does not fully align with the
+  // characteristics of a Buffer.
+  // @see
+  // https://github.com/callstackincubator/react-native-node-api/issues/171
   return napi_create_typedarray(env, ArrayType, length, buffer, 0, result);
 }
 


packages/host/android/CMakeLists.txt
```diff
@@ -20,6 +20,8 @@ add_library(node-api-host SHARED
   ../cpp/Logger.cpp
   ../cpp/CxxNodeApiHostModule.cpp
   ../cpp/WeakNodeApiInjector.cpp
+  ../cpp/node_api_impl.cpp
+  ../cpp/node_api_impl.hpp
```
kraenhansen, line 24: Any reasons we shouldn't follow the existing naming scheme here? Also, I wonder if it makes sense to reference the fact that this is intended for the implementation of the "runtime" part of Node-API?

```suggestion
  ../cpp/RuntimeNodeApi.cpp
  ../cpp/RuntimeNodeApi.hpp
```

packages/node-addon-examples/tests/buffers/addon.js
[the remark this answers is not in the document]
kraenhansen: For now, I think this is a good home for it 🤞

packages/node-addon-examples/tests/buffers/addon.js
```diff
@@ -0,0 +1,31 @@
...
+    .split(",")
+    .map((code) => String.fromCharCode(parseInt(code, 10)))
+    .join("");
+};
+
+console.log(toLocaleString(addon.newExternalBuffer()), addon.theText);
+console.log(addon.newBuffer().toLocaleString(), addon.theText);
+console.log("gc1");
+console.log(addon.getDeleterCallCount(), 0);
+global.gc();
```
kraenhansen, line 17: I don't expect this to be defined on React Native.

packages/node-addon-examples/tests/buffers/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "private": true,
+  "dependencies": {
+    "bindings": "~1.5.0"
+  },
+  "scripts": {
+    "test": "node --expose-gc addon.js"
+  },
+  "gypfile": true,
+  "readme": "ERROR: No README data found!",
+  "_id": "function_arguments@0.0.0"
```
kraenhansen, line 15: What's this? If this was copied from somewhere, we should make sure to attribute accordingly 👍

packages/host/android/CMakeLists.txt
paradowstack: It definitely makes sense to follow existing schema, initially I wanted to follow more node-ish naming convention and put files in separate directory, but if finally they ended up there I will adjust the name accordingly (including `Runtime` part).

packages/host/cpp/node_api_impl.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+napi_status NAPI_CDECL callstack::nodeapihost::napi_create_buffer(
+    napi_env env, size_t length, void** data, napi_value* result) {
+  napi_value buffer;
+  const auto status = napi_create_arraybuffer(env, length, data, &buffer);
+  if (status != napi_ok) {
+    return status;
+  }
+
+  return napi_create_typedarray(env, ArrayType, length, buffer, 0, result);
```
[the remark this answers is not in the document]
paradowstack: It may be worth to mention it, but I am not sure how it should be done. Won't placing a warning on the output clutter the logs? Or we can mention it in the function documentation - but it will be pretty hidden then.
I will create and link the issue to provide an actual implementation.

packages/node-addon-examples/tests/buffers/addon.js
paradowstack: From what I can see Hermes C++ API does not provide a way to trigger a garbage collection. I think without patching the Hermes it won't be possible to run this part of test when running in React Native - please correct me if I am wrong. In this case I will remove it for now.

kraenhansen (approved):
I think it's okay to merge now and look forward to improving the tests with #147.

packages/node-addon-examples/tests/buffers/addon.js
```diff
@@ -0,0 +1,19 @@
...
+    .join("");
+};
+
+console.log(addon.newBuffer().toLocaleString(), addon.theText);
+console.log(toLocaleString(addon.newExternalBuffer()), addon.theText);
+console.log(addon.copyBuffer(), addon.theText);
+let buffer = addon.staticBuffer();
+console.log(addon.bufferHasInstance(buffer), true);
+console.log(addon.bufferInfo(buffer), true);
+addon.invalidObjectAsBuffer({});
```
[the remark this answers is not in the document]
paradowstack: I will update this test addon and add a proper assertion in the next PR.

PR #183, opened 2025-07-16
Added implementation of `V1` async_work runtime functions

paradowstack:
This PR brings in:
- Implementation of following `V1` async work runtime functions using `facebook::react::CallInvoker`:
  - `napi_create_async_work`
  - `napi_queue_async_work`
  - `napi_delete_async_work`
  - `napi_cancel_async_work`
- Introduces test addon, adopted from https://github.com/nodejs/node/blob/main/test/node-api/test_async
- Adds a proper assertion to addon tests
- Fixed an bug in implementation of `napi_create_external_buffer`


packages/host/cpp/RuntimeNodeApiAsync.cpp
```diff
@@ -0,0 +1,183 @@
...
+      current_id_ = 0;
+    }
+    return ++current_id_;
+  }
+
+  IdType current_id_{0};
+  std::unordered_map<IdType, std::shared_ptr<AsyncJob>> jobs_;
+};
+
+static std::weak_ptr<facebook::react::CallInvoker> callInvoker;
```
kraenhansen, line 83: While this likely works in most cases, there could be multiple call invokers in flight, especially during hot-module reloading where the JSI runtime is being torn down and recreated for the "new" app instance.

My instinct would be to store the call invoker in a structure which would be indexable by the `napi_env` to ensure the correct call invoker is used for the correct `napi_env`.

kraenhansen: I would live for this to be addressed in this PR, if you agree and see a way forward. Alternatively, we could merge and release this and create an issue to track an improvement to make it more robust.

kraenhansen (approved):

PR #188, opened 2025-07-17
chore: better gradle setup

artus9033:
This PR introduces the following:
- a new script to the host package (`react-native-node-api`): `consumerSettings.gradle` that shall be included in the demo app's `settings.gradle`; the integration has been described on the bottom of this post & in the `ANDROID.md` docs
- fixes the `hermes.ts` script which set `isEnabled` to `ora` spinners based on the silent mode flag value, which wasn't really a silent mode but just turned off animations of the spinners, by replacing these with `isSilent` to ensure a true silent mode
- adds a `postinstall` script to the root workspace `package.json` to automatically `npm run build` all packages upon install for ease of development

There unfortunately is no way to set env vars from Gradle scripts, therefore automation of export of `REACT_NATIVE_OVERRIDE_HERMES_DIR` is impossible to be achieved. If this is meant to be a temporary setup, then I would leave this step as manual for the consumer; if this is meant to be permanent, I would send a PR to RN, adding optional secondary resolution of custom Hermes path in [those files](https://github.com/search?q=repo%3Afacebook%2Freact-native%20REACT_NATIVE_OVERRIDE_HERMES_DIR&type=code) using Gradle Properties instead, if this is worth the effort. CC @paradowstack @kraenhansen

---

Gradle script integration code to be added to any app's `settings.gradle`:

```
// customize the path to match your node_modules location
apply(from: "../../../node_modules/react-native-node-api/android/consumerSettings.gradle")
applyNodeAPISettings(settings)
```


packages/host/android/consumerSettings.gradle
```diff
@@ -0,0 +1,22 @@
+import org.gradle.initialization.DefaultSettings
+
+
+ext.applyNodeAPISettings = { DefaultSettings settings ->
+    def searchDirectory = rootDir.toPath()
+    do {
+        def p = searchDirectory.resolve("node_modules/react-native")
+        if (p.toFile().exists()) {
+            println "[RN-NAPI] Found React Native in ${p.toRealPath().toString()}"
```
kraenhansen, line 9: As per https://nodejs.medium.com/renaming-n-api-to-node-api-27aa8ca30ed8 we want to respect the Node.js project decision and use "Node-API" instead of "N-API" or "NAPI".

```suggestion
            println "[Node-API] Found React Native in ${p.toRealPath().toString()}"
```

docs/ANDROID.md
```diff
@@ -1,13 +1,36 @@
...
+
+### How it works: building Hermes from source
 
 Because we're using a version of Hermes patched with Node-API support, we need to build React Native from source.
 
+Alternatively, if for whatever reason you want to do it manually, you can do so by exporting this environment variable before Gradle invocation:
+
 ```
-export REACT_NATIVE_OVERRIDE_HERMES_DIR=`npx react-native-node-api vendor-hermes --silent`
+export REACT_NATIVE_OVERRIDE_HERMES_DIR="$(npx react-native-node-api vendor-hermes --silent)"
```
kraenhansen, line 28: 👍

docs/IOS.md
```diff
@@ -0,0 +1,7 @@
+# iOS support
+
+## iOS setup: script for adjusting environment variables
+
+To integrate automatic setup of Hermes engine, a special env variable (`REACT_NATIVE_OVERRIDE_HERMES_DIR`) must be set to a proper path. We provide the script `react-native-node-api vendor-hermes --silent` which will output a single line, the path to Hermes directory.
+
+Each time you run XCode, make sure this is in place.
```
kraenhansen, line 7: This is not true, it only needs to be set when running "pod install" and we do that automatically - no need for manual steps on iOS, not even when running XCode.

```suggestion
```

```suggestion
```

docs/ANDROID.md
```diff
@@ -1,13 +1,36 @@
...
+
+```groovy
+// customize the path to match your node_modules location
+apply(from: "../../../node_modules/react-native-node-api/android/consumerSettings.gradle")
+applyNodeAPISettings(settings)
+```
+
+### Step 2: script for adjusting environment variables
+
+To integrate automatic setup of Hermes engine, a special env variable (`REACT_NATIVE_OVERRIDE_HERMES_DIR`) must be set to a proper path. Since Gradle does not really support loading `.env` files, this must be automated by the consumer. We provide the script `react-native-node-api vendor-hermes --silent` which will output a single line, the path to Hermes directory.
```
kraenhansen, line 17:
```suggestion
A special environment variable (`REACT_NATIVE_OVERRIDE_HERMES_DIR`) must be set to the path of a Hermes engine with Node-API support. Since Gradle does not support loading `.env` files directly, this must be automated by the consumer. We provide the `react-native-node-api vendor-hermes --silent` command, which will download Hermes and output the path to Hermes directory path as its only output.
```

docs/ANDROID.md
```diff
@@ -1,13 +1,36 @@
...
+
+Each time you run Android Studio, make sure this is in place.
```
kraenhansen, line 19:
```suggestion
Each time you run Android Studio or start the Android app from a terminal, make sure this is in place.
```

docs/ANDROID.md
```diff
@@ -1,13 +1,36 @@
...
+
+### How it works: building Hermes from source
 
 Because we're using a version of Hermes patched with Node-API support, we need to build React Native from source.
 
+Alternatively, if for whatever reason you want to do it manually, you can do so by exporting this environment variable before Gradle invocation:
```
kraenhansen, line 25: I don't understand this sentence. Unfortunately, all of this is pretty manual at the moment. Do you mean, if you want to build and run the app via a terminal?

kraenhansen:
First off - thanks a ton for your interest in the project!

I recognize that your PR is in draft and there's a few things that I'd like for you to address before merging this.

docs/ANDROID.md
```diff
@@ -1,13 +1,36 @@
 # Android support
 
-## Building Hermes from source
+## Android setup
```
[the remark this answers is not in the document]
artus9033: Thanks for spotting this.

package.json
```diff
@@ -21,7 +21,8 @@
     "test": "npm run test --workspace react-native-node-api --workspace cmake-rn --workspace gyp-to-cmake --workspace node-addon-examples",
     "bootstrap": "npm run build && npm run bootstrap --workspaces --if-present",
     "prerelease": "npm run build && npm run prerelease --workspaces --if-present",
-    "release": "changeset publish"
+    "release": "changeset publish",
+    "postinstall": "npm run build"
```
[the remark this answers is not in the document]
artus9033: Sure thing, I got rid of that.

package-lock.json
```diff
@@ -33,6 +34,7 @@
     "apps/test-app": {
       "name": "@react-native-node-api/test-app",
       "version": "0.1.1",
+      "hasInstallScript": true,
```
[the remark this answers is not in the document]
artus9033: Good find, seems like a leftover of my experiment in the test app - it's removed now.

docs/ANDROID.md
```diff
@@ -1,13 +1,36 @@
...
+
+### Step 1: `settings.gradle`
+
+Gradle needs special handling to build React Native from source. In your app's `settings.gradle` please include the below:
+
+```groovy
+// customize the path to match your node_modules location
+apply(from: "../../../node_modules/react-native-node-api/android/consumerSettings.gradle")
```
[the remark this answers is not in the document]
artus9033: Right, it may be more appropriate for script names.

docs/ANDROID.md
artus9033: Thanks, this is more clear - applied.

docs/ANDROID.md
artus9033: I'd only change 'start' to 'build' - applied 👍

docs/ANDROID.md
artus9033: Right, this is also a leftover I have to correct.

artus9033:
Sure thing, thank you for the insights @kraenhansen . All we discussed is applied in the latest commit. I left one comment to discuss the test-app's `settings.gradle`

docs/ANDROID.md
```diff
@@ -2,11 +2,32 @@
...
-Because we're using a version of Hermes patched with Node-API support, we need to build React Native from source.
+### Step 1: `settings.gradle`
 
+Gradle needs specific configuration to build React Native from source: dependency substitutions for React Native & Hermes modules, and modifications to default logic in RN build scripts, which is done by setting an environment variable, as described in this section.
+
+In your app's `settings.gradle` please include the below:
+
+```groovy
+// customize the path to match your node_modules location
+apply(from: "../../../node_modules/react-native-node-api/android/app-settings.gradle")
```
kraenhansen, line 13: When we introduce our own settings script it becomes a bit more magical IMO and I'm not 100% convinced this is an increase in DX 🤔 There's something "simple" in just instructing users to build React Native from source and defer to whatever upstream documentation is available for doing that, with a code-snippet to make it easier.

docs/ANDROID.md
```diff
@@ -2,11 +2,32 @@
...
+applyNodeAPISettings(settings)
 ```
-export REACT_NATIVE_OVERRIDE_HERMES_DIR=`npx react-native-node-api vendor-hermes --silent`
+
+### Step 2: script for adjusting environment variables
+
+> [!IMPORTANT]
+> Each time you run Android Studio or build the Android app from a terminal, make sure the below is in place.
+
+Because we're using a version of Hermes patched with Node-API support, we need to build React Native from source. A special environment variable (`REACT_NATIVE_OVERRIDE_HERMES_DIR`) must be set to the path of a Hermes engine with Node-API support. Since Gradle does not support loading `.env` files directly, this must be automated by the consumer. We provide the `react-native-node-api vendor-hermes --silent` command, which will download Hermes and output the path to Hermes directory path as its only output.
```
kraenhansen, line 22:
```suggestion
Because we're using a version of Hermes patched with Node-API support, we need to build React Native and Hermes from source. A special environment variable (`REACT_NATIVE_OVERRIDE_HERMES_DIR`) must be set to the path of a Hermes engine with Node-API support. Since Gradle does not support loading `.env` files directly, this must be automated by the app developer. We provide the `react-native-node-api vendor-hermes --silent` command, which will download Hermes and output the path to Hermes directory path as its only output.
```

docs/ANDROID.md
```diff
@@ -2,11 +2,32 @@
...
+
+You can configure the environment variable using the following command:
+
 ```
+export REACT_NATIVE_OVERRIDE_HERMES_DIR="$(npx react-native-node-api vendor-hermes --silent)"
+```
+
+This either needs to be done each time before Gradle / Android Studio invocation, or permanently in a shell init script such as `~/.zshrc` on Zsh (MacOS) or `~/.bashrc` on Bash (Linux).
```
kraenhansen, line 30: I'm not sure this is a good suggestion. It has the potential to mess up people's setup, quite significantly - in a way which is really hard to debug.

kraenhansen (commented):
Code seems fine 👍 I'm curious if this is actually an improvement in DX and would like a bit more feedback before merging, if that's okay?

docs/ANDROID.md
shirakaba, line 13: This path (`../../../node_modules`) assumes a monorepo structure, so won't work on a regular repo.

Can we use proper Node module resolution instead? As demonstrated [here](https://github.com/expo/expo/blob/ab9f960751807743c596f80b56e698118d1aeb53/templates/expo-template-bare-minimum/android/app/build.gradle#L13).

shirakaba: Right now we have asymmetry between the iOS and Android sides – iOS is set up automatically, while Gradle isn't. So I think we need to at least make things consistent.

If this were an Expo project, I'd probably arrange it as a config plugin called `withBuildReactNativeFromSource`, so that it involves no magic specific `react-native-node-api` and is just clearly a convenience function.

**I think it's okay to defer to the upstream docs for building React Native from source. Maybe that part needn't be handled by our plugin, then?**

Something I would like to improve the DX on, though, is this part:

```sh
export REACT_NATIVE_OVERRIDE_HERMES_DIR=`npx react-native-node-api vendor-hermes --silent`
```

It's a pain because it relies on shell stuff, so there's great potential for people to get it wrong (e.g. run it in the wrong directory or forget to export it again).

If this were an Expo project, I'd probably say that `npx react-native-node-api vendor-hermes` should just append the necessary env var into `.env.local` for you. But because React Native Community CLI projects don't have any standard env file that they source variables from, and because React Native strictly grabs that variable via `System.getenv()` (in both [react-native-gradle-plugin](https://github.com/facebook/react-native/blob/52f2cbab41297d4e5689c112188f67d3c6150fe9/packages/gradle-plugin/react-native-gradle-plugin/src/main/kotlin/com/facebook/react/utils/PathUtils.kt#L140) and [hermes-engine](https://github.com/facebook/react-native/blob/52f2cbab41297d4e5689c112188f67d3c6150fe9/packages/react-native/ReactAndroid/hermes-engine/build.gradle.kts#L65)), I'm not sure we have any great options.

Any ideas?

shirakaba (commented):
Have added some thoughts!

And thanks @artus9033, I was worrying about the DX on this part so this is a great PR to see. Let's keep talking about the ideal shape it should take.

kraenhansen:
Revisiting this while checking whether the manual `vendor-hermes` step is still needed on Android. Findings, so the PR isn't blocking on a stale picture:

**The Gradle settings helper is still a valid idea, but it's no longer needed by this repo.** `react-native-test-app` now applies exactly these four dependency substitutions itself when `react.buildFromSource=true` is set (`test-app.gradle`), which is what `apps/test-app/android/gradle.properties` does — so our own test app builds React Native from source without touching `settings.gradle`. Apps not based on `react-native-test-app` still hand-edit it, so a consumer-side script would still save them the boilerplate.

**The environment variable is still unavoidable.** React Native reads the override with `System.getenv("REACT_NATIVE_OVERRIDE_HERMES_DIR")` in `ReactAndroid/hermes-engine/build.gradle.kts`, and Gradle can't set a variable for its own build — so the manual export stands, exactly as this PR concluded. Vendoring itself is also still required after #372: we now build Hermes from a pinned `static_h` commit for its first-party Node-API rather than patching it, and the host links `hermes_napi_create_env` from it.

**Two parts of the diff have been overtaken:**

- The `hermes.ts` hunks: #372 rewrote that command (the JSI-copy step is gone), but the `isEnabled: !silent` → `isSilent` fix was real — a disabled ora spinner still writes its text and final symbol to stderr. Picked into #410.
- The docs and `.nvmrc`: the repo is on Node 24 now, and the Hermes wording changed with #372. `docs/ANDROID.md` has been reworked in #411, which also documents the `react.buildFromSource=true` route and why the env var has to be exported per shell.

Leaving this open for now — the remaining piece is whether we want to ship a `settings.gradle` helper for plain React Native apps. It would need a rebase onto `next` (this targets `main`, and currently conflicts).

---

PR #193, opened 2025-07-23
Implementation of `napi_fatal_error` and `napi_get*_version` functions

paradowstack:
This pull request introduces enhancements to the logging system and  adds new Node-API functions.

### Logging Enhancements:
* Refactored the logging system to support multiple log levels (`Debug`, `Warning`, `Error`)
### Node-API Functionality:
* Added `napi_fatal_error` to log fatal errors and terminate the process
* Introduced `napi_get_node_version` and `napi_get_version` functions to provide versioning information for Node.js and Node-API

### DIFF
--- a/packages/host/cpp/Logger.cpp
+++ b/packages/host/cpp/Logger.cpp
@@ -9,27 +9,77 @@
 #include <TargetConditionals.h>
 #endif
 
-namespace callstack::nodeapihost {
-void log_debug(const char *format, ...) {
-  // TODO: Disable logging in release builds
+namespace {
+constexpr auto LineFormat = "[%s] [NodeApiHost] ";
 
-  va_list args;
-  va_start(args, format);
+enum class LogLevel { Debug, Warning, Error };
+
+constexpr std::string_view levelToString(LogLevel level) {
+  switch (level) {
+    case LogLevel::Debug:
+      return "DEBUG";
+    case LogLevel::Warning:
+      return "WARNING";
+    case LogLevel::Error:
+      return "ERROR";
+    default:
+      return "UNKNOWN";
+  }
+}
 
 #if defined(__ANDROID__)
-  __android_log_vprint(ANDROID_LOG_DEBUG, LOG_TAG, format, args);
+constexpr int androidLogLevel(LogLevel level) {
+  switch (level) {
+    case LogLevel::Debug:
+      return ANDROID_LOG_DEBUG;
+    case LogLevel::Warning:
+      return ANDROID_LOG_WARN;
+    case LogLevel::Error:
+      return ANDROID_LOG_ERROR;
+    default:
+      return ANDROID_LOG_UNKNOWN;
+  }
+}
+#endif
+
+void log_message_internal(LogLevel level, const char* format, va_list args) {
+#if defined(__ANDROID__)
+  __android_log_vprint(androidLogLevel(level), LOG_TAG, format, args);
 #elif defined(__APPLE__)
   // iOS or macOS
-  fprintf(stderr, "[NodeApiHost] ");
+  const auto level_str = levelToString(level);
+  fprintf(stderr, LineFormat, level_str.data());
   vfprintf(stderr, format, args);
   fprintf(stderr, "\n");
 #else
   // Fallback for other platforms
-  fprintf(stdout, "[NodeApiHost] ");
+  const auto level_str = levelToString(level);
+  fprintf(stdout, LineFormat, level_str.data());
   vfprintf(stdout, format, args);
   fprintf(stdout, "\n");
 #endif
+}
+}  // anonymous namespace
 
+namespace callstack::nodeapihost {
+
+void log_debug(const char* format, ...) {
+  // TODO: Disable logging in release builds
+  va_list args;
+  va_start(args, format);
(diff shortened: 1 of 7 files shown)


packages/host/cpp/Versions.hpp
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#define NODE_MAJOR_VERSION 8
+#define NODE_MINOR_VERSION 6
+#define NODE_PATCH_VERSION 0
+
+#define NAPI_VERSION 1
```
paradowstack, line 7: I set it to the lower possible versions - I am happy to change that.
I think that `NAPI_VERSION` should remain `1`, but what about `NODE_` versions?

kraenhansen (approved):
Small comments that could come as follow-ups.

packages/host/cpp/Versions.hpp
paradowstack: > All of the https://github.com/nodejs/node-api-headers/blob/main/include%2Fjs_native_api.h functions are implemented to `NAPI_VERSION=10` and all of the runtime functions does have failing implementations provided by the weak-node-api binary, so I'd think we could safely bump this to 10 🤔 In fear that existing add-ons might fail to use js-native-apis from Hermes of it's set to 1.

I think that functions are implemented up to `NAPI_VERSION=8` in [node-api-headers](https://github.com/nodejs/node-api-headers/blob/d46edabd5bbacc08393578eb19fc2eba7021b597/include/js_native_api.h#L17) and [hermes_node_api does not provide implementations for 9 and 10 versions ](https://github.com/vmoroz/hermes-windows/blob/8c054abbc17605cae32a5e77255b5e8f5a355be7/API/hermes_node_api/hermes_node_api.cpp#L6813) - so we can bump it to `8`? Please correct me if I understood it wrongly 🤔

packages/host/cpp/RuntimeNodeApi.cpp
```diff
@@ -121,4 +123,40 @@ napi_status napi_create_external_buffer(napi_env env,
   return napi_create_typedarray(env, ArrayType, length, buffer, 0, result);
 }
 
+void napi_fatal_error(const char* location,
+    size_t location_len,
+    const char* message,
+    size_t message_len) {
+  log_error("Fatal error in Node-API: %.*s: %.*s",
```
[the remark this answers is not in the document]
paradowstack: > Perhaps the namespace is sufficient already?
> 
> alternatively

I would probably combine your suggestions and go with:
`log_error("Fatal Node-API error: %.*s (in %.*s)` 🤔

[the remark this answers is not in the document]
paradowstack: > This is the reference implementation: https://github.com/nodejs/node/blob/6bb08f7f8e4c984eeccbd0692410cd88ebc2b528/src/node_errors.cc#L574
> 
> I see that it's flushing `stderr`, perhaps we'd want to do the same somehow? 🤔

Well explicitly flushing `stderr` is not required because `stderr` is unbuffered (unlike `stdout`), and output is sent immediately so I think we don't need that 🤔

packages/host/cpp/Logger.cpp
```diff
@@ -9,27 +9,77 @@
...
   fprintf(stderr, "\n");
 #else
   // Fallback for other platforms
-  fprintf(stdout, "[NodeApiHost] ");
+  const auto level_str = levelToString(level);
+  fprintf(stdout, LineFormat, level_str.data());
   vfprintf(stdout, format, args);
   fprintf(stdout, "\n");
 #endif
+}
```
shirakaba, line 61: This logs to the native side - totally a good starting point, but would it make sense in future to log to the JS console so that we see it in the Metro/Chrome logs?

paradowstack: Overall, it makes sense, but is the appearance of native logs in the JS console something common? 🤔

shirakaba: Hard to say because Node.js is the main precedent here and I imagine these logs show up in its logs.

Mainly, if the Node-API module logs a fatal error, it feels like it might be good to be able to see it without having to attach Xcode/LLDB/Android Studio to the process.

kraenhansen: @shirakaba I agree it would be great to make these more discoverable, although I expect them to be rare. I'm a bit concerned though, that it might not be possible because of the semantics of the fatal error / exception functions, which cannot return and must abort the process. There won't be an app to display the error nor a DevTools connection to transport it once the program leaves the function.

paradowstack: Could we put a line to the JS console just before we abort the process, so it is the last visible line and gives the developer information what went wrong?

kraenhansen: I'm a bit unsure on the semantics which need to be upheld by this function. If we call into other Node-API functions, to propagate the log message to JS, we might potentially (an edge-case for sure) get into an endless recursion. And hinted above, even if we called into `console.log` as an example, it won't make it into the DevTools client because the network transporting it from the app is likely not blocking and the app will die before it gets to send the message.

That's my hunch - we could definitely experiment with it 👍

PR #200, opened 2025-07-24
feat: add generate project + project improvements

dougg0k:
- [ ] Created `generate` command to generate scaffold project. Replacing ferric-example.
- [x] Add `path` as arg to build project.
- [x] Fixed cargo build command with correct `--release` flag.
- [x] Added `Cargo.toml` optmizations for size.
- [ ] Added `tests` to ferric-cli, to build and generate commands.
- [ ] Changed `verbose` flag to also log commands.
- [x] Replaced `eslint` since the project lacks `formatter` and `sort organizer`, `biome` is faster and has all three.
- [x] Changed default build output to send to dist folder.

### DIFF
--- a/packages/ferric/README.md
+++ b/packages/ferric/README.md
@@ -27,4 +27,9 @@ lib_name
 │   .gitignore
 └───src
 │   │   lib.rs
+└───dist (after build)
+│   │   index.js
+│   │   index.d.ts
+│   │   native_android_folder
+│   │   native_ios_folder
 ```


kraenhansen:
Woah - hold that 🐴 While I really appreciate your enthusiasm, I'd love for your contributions to:
1. Address specific issues, which we can discuss before jumping on a specific solution (feel free to create these issues yourselves, as you're discovering stuff which could be improved)
2. keep your PRs separate such that they each address just one issue (or multiple if they're truly related / duplicates).

You're making a lot of suggested changes here that I'm not comfortable with and I'd hate for good suggestions to be held back because of some disagreement on what tool to use for linting (as an example).

.github/workflows/check.yml
```diff
@@ -9,7 +9,7 @@ on:
     branches:
       - main
   pull_request:
-    types: [opened, synchronize, reopened]
```
kraenhansen, line 12: Why?

packages/ferric-example/package.json
```diff
@@ -1,21 +0,0 @@
-{
-  "name": "@react-native-node-api/ferric-example",
```
kraenhansen, line 2: I don't understand the need to remove this in its entirety? While I do agree having a command to generate a project scaffolding, this package serves a purpose as a point of reference in itself.

dougg0k: I was going to try and replace that with tests.

dougg0k: If the cli are generating the project, then that's the point of reference.

.github/workflows/check.yml
dougg0k: I didnt remove that line, I replaced one type with another.

biome.json
```diff
@@ -0,0 +1,38 @@
+{
+  "$schema": "https://biomejs.dev/schemas/2.1.2/schema.json",
```
kraenhansen, line 2: Iff we'd switch linter, it would have to be in a completely separate PR without any added features, and we'd have to bike-shed that for a while on Discord first.

.github/workflows/check.yml
kraenhansen: As per https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=synchronize#pull_request

The "synchronize" event that you suggest removing ensures we trigger a build when the main branch gets an update which might break the PR. I see no reason for removing that.

> A pull request's head branch was updated. For example, the head branch was updated from the base branch or new commits were pushed to the head branch.

packages/ferric-example/package.json
kraenhansen: > that's the point of reference.

It's hard to share a link to something which comes out of a CLI.

dougg0k: https://github.com/callstackincubator/react-native-node-api/pull/200/commits/138fd49cbc31a82fdb7f58d83ee11ae21b510531

biome.json
kraenhansen: > Any reason why?

I added a few comments here or there, but I won't be able to give a thorough review if your actual changes are interleaved with a ton of changes from replacing the linter, that's an unreasonable ask IMO.

dougg0k: Maybe it's better to close the PR then?

packages/ferric-example/package.json
dougg0k: It could, but only after generated through the testing.

dougg0k: I meant mostly as to keep the whole process cleaner.

biome.json
kraenhansen: I don't want to discourage you by blatantly closing the PR - I'll let you decide its faith.

I would love to discuss and merge many of the improvements, if we're able to get them reviewed one-by-one 👍

PR #202, opened 2025-07-24
chore: replace linter

dougg0k:
Replaced eslint with biomejs.

Eslint only has linting where biome has lint, formatter and sort organizer. A much more complete tool.

Previous discussion here https://github.com/callstackincubator/react-native-node-api/pull/200/files#r2229485568

A bunch of linting and formatting fixes.


shirakaba:
At work, we have been stuck on Biome for the last year and this week finally migrated off it back to Prettier + Eslint. Main reasons were:

- Biome's React linting rules are less powerful (e.g. it can only tell a ref is stable if that ref is declared in the same scope).
- There are many cases where you have to manually write Biome ignores, because the tooling does not support inserting them via a keyboard shortcut.
- It co-exists horribly with Prettier. I always have to turn it on as my default formatter in my VS Code user settings, which prevents me from jumping between Prettier and Biome projects.
- It can't handle a Windows VM running on macOS. It installs the wrong target triplet for the Biome executable, so you need to provide an explicit absolute path to it, which is only specific to your environment, and so that setting can't ever be saved into source-control as it'd affect your teammates.
- The type-aware linting rules are very early-stage.
- Prettier is now actually getting faster, with [oxc and the new experimental CLI](https://prettier.io/blog/2025/06/23/3.6.0).

The PR itself looks perfectly correct (though I'd encourage configuring the trailing comma settings to reduce the diff), I just have had a bad experience with Biome and if the hard work of setting up the repo for Eslint has already been done, I don't see much to gain from swapping the linter.

dougg0k:
Here I stopped using prettier, even in vscode, too clunky.

kraenhansen:
@dougg0k it will likely hang around for a bit as I'd like to circle this suggestion with other stakeholders of the project too and spend some time with the branch myself. I'm mostly offline for about another week of vacationing 🏝️ and it likely won't merge before I get back, if at all.

dougg0k:
Project lacks a formatter currently, ideally this would be merged, if someone has better impl, they could change after. @shirakaba

kraenhansen:
I agree we should add a formatter and I opened this PR ~ 2 months ago that I'd like for you to review: #115. It was stalled because we had many open PRs and merging it would cause a lot of conflicts. We still have a bunch, but it might be a good time to revive it.

I surfaced this PR with the rest of the project stakeholders and we're happy with an eslint + prettier setup for now. I agree we do need to tune the rules and plugins further and would love suggestions on that - ideally we'd want to stick pretty close to recommended settings for now.

dougg0k:
Maybe merge that then.

PR #204, opened 2025-07-25
chore: migrated to package manager to pnpm

dougg0k:
Before I had trouble using `npm i`, even if it were recognized in the package-lock as some package as installed, when running commands in a folder, it would not work, unless I specifically ran install (`non-npm`).

This led to this PR. Which in fact work now just after a `pnpm install` in the root folder.


shirakaba:
> Before I had trouble using npm i, even if it were recognized in the package-lock as some package as installed, when running commands in a folder, it would not work, unless I specifically ran install (non-npm).

I didn't experience that when I cloned and installed the repo. Would it help if we just regenerated the `package-lock.json`?

What version of `npm` are you using? I'm using 10.9.2.

dougg0k:
Same version.

dougg0k:
```
ackages/node-addon-examples test: node:internal/fs/promises:955
packages/node-addon-examples test:   const result = await PromisePrototypeThen(
packages/node-addon-examples test:                  ^
packages/node-addon-examples test: Error: ENOENT: no such file or directory, scandir '/home/runner/work/react-native-node-api/react-native-node-api/packages/node-addon-examples/examples/5-async-work/async_work_thread_safe_function/napi/build/Release/binding.android.node/armeabi-v7a'
packages/node-addon-examples test:     at async Object.readdir (node:internal/fs/promises:955:18)
packages/node-addon-examples test:     at async verifyAndroidPrebuild (/home/runner/work/react-native-node-api/react-native-node-api/packages/node-addon-examples/scripts/verify-prebuilds.mts:28:24)
packages/node-addon-examples test:     at async <anonymous> (/home/runner/work/react-native-node-api/react-native-node-api/packages/node-addon-examples/scripts/verify-prebuilds.mts:90:5) {
```

dougg0k:
@shirakaba Can you review? In some fixes, you should consider the comments I made. It's related.

packages/host/package.json
```diff
@@ -88,6 +88,7 @@
   "devDependencies": {
     "@babel/core": "^7.26.10",
     "@babel/types": "^7.27.0",
+    "@types/babel__core": "^7.20.5",
```
kraenhansen, line 91: What was this fixing?

dougg0k: https://github.com/callstackincubator/react-native-node-api/pull/204#issuecomment-3116203736

packages/node-addon-examples/scripts/cmake-projects.mts
```diff
@@ -7,6 +7,9 @@ export const DIRS = [EXAMPLES_DIR, TESTS_DIR];
 
 export function findCMakeProjectsRecursively(dir: string): string[] {
   let results: string[] = [];
+  if (!existsSync(dir)) {
+    return results;
+  }
```
kraenhansen, line 12: This breaks an invariant and shouldn't fail silently.

packages/node-addon-examples/scripts/verify-prebuilds.mts
```diff
@@ -25,6 +25,9 @@ async function verifyAndroidPrebuild(dirent: fs.Dirent) {
   );
   for (const arch of EXPECTED_ANDROID_ARCHS) {
     const archDir = path.join(dirent.parentPath, dirent.name, arch);
+    if (!fs.existsSync(archDir)) {
+      return;
+    }
```
kraenhansen, line 30: This breaks an invariant and shouldn't fail silently.

packages/node-addon-examples/scripts/verify-prebuilds.mts
```diff
@@ -41,6 +44,9 @@ async function verifyApplePrebuild(dirent: fs.Dirent) {
   console.log("Verifying Apple prebuild", dirent.name, "in", dirent.parentPath);
   for (const arch of EXPECTED_XCFRAMEWORK_PLATFORMS) {
     const archDir = path.join(dirent.parentPath, dirent.name, arch);
+    if (!fs.existsSync(archDir)) {
+      return;
+    }
```
kraenhansen, line 49: This breaks an invariant and shouldn't fail silently.

packages/node-addon-examples/scripts/verify-prebuilds.mts
dougg0k: https://github.com/callstackincubator/react-native-node-api/pull/204#issuecomment-3117912710

packages/node-addon-examples/scripts/verify-prebuilds.mts
dougg0k: https://github.com/callstackincubator/react-native-node-api/pull/204#issuecomment-3117912710

kraenhansen:
> Weird that you do not prefer better tooling.

Not weird at all - we're the ones maintaining this package in the long run and choosing technology / tools aren't soly about choosing the seemingly superior solution in my experience. To be clear, I'm not saying that NPM is flawless nor that we won't make a switch to a different package manager and I'm not looking to get more into that conversation right now.

> I gave the reasons in the PR.

I disagree, as I'm unable to reproduce the issue: You're not providing specific commands and outputs from running those and therefore I'm unable to get a strong signal, giving me a compelling reason to accept your suggested change.

dougg0k:
Of course.

I did provide, there should be plenty of commands explaining in the first post and then in the ones that was triggered in the CI.

packages/node-addon-examples/scripts/verify-prebuilds.mts
dougg0k: What would you prefer to be done?

dougg0k:
Another issue appeared here. Which was not generating the ferric build. I then removed node_modules and package-lock.json and reinstalled with npm.

But now the error are this one.

```
...
> cmake-rn --no-auto-link --no-weak-node-api-linkage --xcframework-extension --source ./weak-node-api --out ./weak-node-api

node:internal/modules/package_json_reader:256
  throw new ERR_MODULE_NOT_FOUND(packageName, fileURLToPath(base), null);
        ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'stdin-discarder' imported from /react-native-node-api/packages/cmake-rn/node_modules/ora/index.js
    at Object.getPackageJSONURL (node:internal/modules/package_json_reader:256:9)
    at packageResolve (node:internal/modules/esm/resolve:768:81)
    at moduleResolve (node:internal/modules/esm/resolve:854:18)
    at defaultResolve (node:internal/modules/esm/resolve:984:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:780:12)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:704:25)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:687:38)
    at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:305:38)
    at ModuleJob._link (node:internal/modules/esm/module_job:175:49) {
  code: 'ERR_MODULE_NOT_FOUND'
}
```

dougg0k:
No, with that, it made it worse, as it gave the error mentioned.

dougg0k:
I removed the checks, as previous mentioned in the review, where I had added to prevent the fails, but there are different errors happening, not sure what the approach to fix them.

dougg0k:
From this date fixes, error are back to what they were, where I tried to add a check that wasnt approved due to reasons.

dougg0k:
@kraenhansen Given your last message, I answered but no response were given, should I assume that this PR is not wanted / needed?

Issue #195, opened 2025-07-23
Add path support in ferric-cli


dougg0k:
Hi,

Could you add path arg support in `ferric-cli`? 

As to specify a path of a separated folder inside the project, rather than have multiple package..json.

Thanks.

dougg0k:
Yes, I suppose that would work. Thanks.

PR #205, opened 2025-07-25
feat: improve ferric build

dougg0k:
- [ ] Add `path` as arg to build project. https://github.com/callstackincubator/react-native-node-api/issues/195
- [ ] Changed default build output to send to `dist` folder.
- [ ] Added build `tests`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4020,9 +4020,9 @@
       }
     },
     "node_modules/@napi-rs/cli": {
-      "version": "3.0.3",
-      "resolved": "https://registry.npmjs.org/@napi-rs/cli/-/cli-3.0.3.tgz",
-      "integrity": "sha512-5PxtOcax9PoOriYWYf8BFNgOzUnsogYUDRM8p2leT2VtG/NTMZCyoRQMzSMB0qEaw+Durkefc5jXCpEF6wrWbA==",
+      "version": "3.0.4",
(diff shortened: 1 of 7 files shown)


dougg0k:
All these formatting are due to the project lacking a formatter. So, it formatted by default configs.

dougg0k:
It's clean, should be easy to see the changes. As there isnt that much changes.

dougg0k:
@kraenhansen @shirakaba Can you review, possibly merge and release a new version?

kraenhansen:
I'm currently vacationing (you probably didn't notice my status, which is fine). Please also be mindful that we have other priorities that we're working towards, which this is detracting attention from.

I'm checking my notifications regularly and there's no need to at-mention to get attention - you have it and I'll get to your suggestions and release those when I'm available to handle any issues that may arise from that.

As to your suggested changes:
- there's 100+ lines with changes not related to the core of the issue you're solving
- you're both providing a changeset
- you're addressing multiple improvements in a single PR and one (storing output in `dist`) that doesn't have an issue associated, hence we didn't discuss. I'm honestly not sure if that's what we want to do, so let's discuss that on an issue describing the problem you're trying to solve first.

I need you to rectify all of the above before I consider the PR ready.

dougg0k:
I didnt, github should make that more visible. Sure.

- It should be fine now, just ignore the imports sorting.
- My bad, added.
- Before output files were mixed with source code of the library, so hence why dist/.

Issue #196, opened 2025-07-23
Reduce native files size


dougg0k:
Hi,

I was testing out `ferric-cli`, just with the `ferric-example` it seems to generate a big `.so` file just for android. In this case it were `6.1mb`.

I suppose this can negatively affect the final build size of RN apps.

Is there a way or if possible, could you add rust flags to the CLI, or whatever that it can be done to help reduce size of the native generated file(s), `.so` or otherwise?

https://github.com/johnthagen/min-sized-rust

Thanks.

kraenhansen:
I'd be happy to add anything to reduce the file size - got any concrete ideas in mind?

Notice that the current default `--configuration` is "Debug" and you'll likely get a smaller binary if you pass `--configuration Release` to ferric.

I know that there are also possibilities of shaving off size with link-time optimization (LTO) when building the final app.

dougg0k:
I suggested the repo which already should contain all the possible ways. Whichever you choose. 

Maybe leave `Release` as default?

dougg0k:
Maybe `ferric-example` should `not` exist. But be generated by the `ferric-cli`, that way Cargo.toml and others files could be better managed.

dougg0k:
If I am able to finish this PR, there should be plenty of improvements added.

https://github.com/callstackincubator/react-native-node-api/pull/200 - List inside.

dougg0k:
In the init/generate command, do you prefer that it provide a package.json to be a standalone lib or can it be as the idea I had, you suggest the scripts as in the previous comment, then that package import the lib from the dist/ in the actual project? 

https://github.com/callstackincubator/react-native-node-api/pull/206

@shirakaba @kraenhansen

shirakaba:
> In the init/generate command, do you prefer that it provide a package.json to be a standalone lib or can it be as the idea I had, you suggest the scripts as in the previous comment, then that package import the lib from the dist/ in the actual project?

Sorry, not quite clear what you mean. Could you illustrate the two alternatives?

dougg0k:
1- Not a standalone package, without package.json. But, managed by the package.json root of the app.

```markdown
lib_name
│   Cargo.toml
│   build.rs
│   .gitignore
└───src
│   │   lib.rs
└───dist (after build)
│   │   index.js
│   │   index.d.ts
│   │   native_android_folder
│   │   native_ios_folder
```

2- With a package.json to be a standalone lib.

dougg0k:
@shirakaba 

[...]

PR #206, opened 2025-07-25
feat: add generate command

dougg0k:
- [ ] Created `generate` command to generate / scaffold the project. 
- [ ] Replacing `ferric-example` with `app-test`, scaffolding the project there.
- [ ] Add tests.
- [x] Added `Cargo.toml` optmizations for size. https://github.com/callstackincubator/react-native-node-api/issues/196


packages/ferric/src/template.ts
```diff
@@ -0,0 +1,58 @@
...
+
+[build-dependencies]
+napi-build = "2"
+
+[profile.release]
+lto = true
+codegen-units = 1
+strip = "symbols"
+opt-level = "z"
+panic = "abort"
```
shirakaba, line 30: Although this does make a small release build, some of these optimisations have big consequences (e.g. `panic = "abort"`) so I probably wouldn't make them defaults.

packages/ferric/src/template.ts
```diff
@@ -0,0 +1,58 @@
+export function cargoTomlTemplate(projectName: string) {
```
shirakaba, line 1: May be preferable to have the `Cargo.toml` template file as a standalone file that the CLI reads in, so that we get proper syntax highlighting. It can have some template strings like `"PROJECT_NAME_REPLACE_ME"`.

packages/ferric/src/template.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+`;
+
+export const LIB_RS_TEMPLATE = `
+use napi_derive::napi;
+
+#[napi]
+pub fn sum(a: i32, b: i32) -> i32 {
+    a + b
+}
```
shirakaba, line 57: Again, may be preferable to have these as standalone files that the CLI reads in, so that we get proper syntax highlighting.

packages/ferric/src/template.ts
shirakaba, line 1: I wonder whether the CLI should rather take the up-to-date standard Cargo library template and modify it, rather than having a template that may go out of date over time. There are pros and cons either way, of course.

packages/ferric-example/package.json
shirakaba, line 1: I think it's still handy to keep the `ferric-example` around for reference (and development) even if adding a `generate` command. Let's not delete this.

packages/ferric/package.json
```diff
@@ -13,15 +13,19 @@
...
     "bufout": "^0.3.2",
     "chalk": "^5.4.1",
     "commander": "^13.1.0",
-    "react-native-node-api": "0.3.2",
-    "ora": "^8.2.0"
+    "ora": "^8.2.0",
+    "react-native-node-api": "0.3.2"
+  },
+  "devDependencies": {
+    "vitest": "^3.2.4"
```
shirakaba, line 29: Could we take out the tests (and addition of `vitest`) from this PR to simplify the review? Feels more like they belong in an independent PR.

shirakaba (commented):
Some initial thoughts. I'm not too familiar with the Ferric stuff, so I defer to others.

packages/ferric/src/template.ts
dougg0k: Unlikely that will keep changing. Better to have it predefined. Only thing that can most likely change are the edition year.

packages/ferric/src/template.ts
dougg0k: Ok

packages/ferric/src/templates/Cargo.toml
```diff
@@ -11,7 +11,7 @@ crate-type = ["cdylib"]
 version = "3.1"
 default-features = false
 # see https://nodejs.org/api/n-api.html#node-api-version-matrix
-features = ["napi3"]
+features = ["napi4"]
```
kraenhansen, line 14: Last time I checked, this wasn't possible. If I remember correctly, it was due to the missing implementation of the Node-API threadsafe functions 🤔 But I might be wrong - that needs to be investigated for this to merge.

packages/ferric/package.json
kraenhansen: Please be mindful of the technology choices already made in the project and help keep things converged. 

We're using the Node.js built-in test runner via tsx (`tsx --test`) in the other packages. (And I'm not looking for a debate on which is the superior choice in a PR review comment 🙈)

packages/ferric/src/templates/Cargo.toml
dougg0k: Hmm.

dougg0k:
I see, didnt know.

dougg0k:
Ready for review.

dougg0k:
@kraenhansen

PR #232, opened 2025-09-13
Enable `RCT_USE_RN_DEP` in test app and assert `RCT_USE_PREBUILT_RNCORE` != `1`

kraenhansen:
As an alternative to https://github.com/callstackincubator/react-native-node-api/pull/207 want the test app to enable `RCT_USE_RN_DEP` (since we can and this is faster) and assert that `RCT_USE_PREBUILT_RNCORE` is never enabled (since we cannot support this as the prebuild of React Native Core use a JSI header which we have to patch - resulting in runtime crashes when enabled with our patched Hermes).

### DIFF
--- a/packages/host/react-native-node-api.podspec
+++ b/packages/host/react-native-node-api.podspec
@@ -41,26 +41,11 @@ Pod::Spec.new do |s|
     CMD
   }
 
-  # Use install_modules_dependencies helper to install the dependencies if React Native version >=0.71.0.
+  # Use install_modules_dependencies helper to install the dependencies (requires React Native version >=0.71.0).
   # See https://github.com/facebook/react-native/blob/febf6b7f33fdb4904669f99d795eba4c0f95d7bf/scripts/cocoapods/new_architecture.rb#L79.
   if respond_to?(:install_modules_dependencies, true)
     install_modules_dependencies(s)
   else
-    s.dependency "React-Core"
-    # Don't install the dependencies when we run `pod install` in the old architecture.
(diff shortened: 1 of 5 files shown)


shirakaba (approved):

kraenhansen:
Converted this into draft: I don't think we'll actually want to merge and add this matrix. We'll likely just pick one configuration (most likely `RCT_USE_PREBUILT_RNCORE=1`) and run the tests with that.

packages/host/react-native-node-api.podspec
```diff
@@ -41,26 +41,11 @@ Pod::Spec.new do |s|
...
-          "HEADER_SEARCH_PATHS" => "\"$(PODS_ROOT)/boost\"",
-          "OTHER_CPLUSPLUSFLAGS" => "-DFOLLY_NO_CONFIG -DFOLLY_MOBILE=1 -DFOLLY_USE_LIBCPP=1",
-          "CLANG_CXX_LANGUAGE_STANDARD" => "c++17"
-      }
-      s.dependency "React-Codegen"
-      s.dependency "RCT-Folly"
-      s.dependency "RCTRequired"
-      s.dependency "RCTTypeSafety"
-      s.dependency "ReactCommon/turbomodule/core"
-    end
```
kraenhansen, line 64: Drive-by deleting this as we don't support old versions of React Native anyway.

kraenhansen: An error is raised:
> This version of React Native is too old for React Native Node-API

Issue #163, opened 2025-07-08
Sanity check of Hermes source directory via environment variable, when building Android projects


kraenhansen:
I've investigated ways of letting the host package drive building Hermes from (our) source. So far, I haven't found any.

In the absence of a fully automated solution, it would be good if we exist early with instructions on how to set the `REACT_NATIVE_OVERRIDE_HERMES_DIR` environment variable if missing:

```
export REACT_NATIVE_OVERRIDE_HERMES_DIR=`npx react-native-node-api vendor-hermes --silent`
```

I would expect the error to be thrown from a new task around here: 

https://github.com/callstackincubator/react-native-node-api/blob/8dac80570cea5c9790b3bcc0a84237f59cd0d4d6/packages/host/android/build.gradle#L141-L157

huytdps13400:
This is already implemented on the current `next` branch. At head `0a29fbd`, `packages/host/android/build.gradle` fails immediately when `REACT_NATIVE_OVERRIDE_HERMES_DIR` is missing and prints the exact setup command requested here:

```sh
export REACT_NATIVE_OVERRIDE_HERMES_DIR=$(npx react-native-node-api vendor-hermes --silent)
```

It also links to the React Native build-from-source guide. No additional PR is needed for this issue; it can be closed.

PR #234, opened 2025-09-14
Check Hermes override pre-condition from Host package when building for Android

kraenhansen:
Ideally, we would have liked to inject the overriding of Hermes source directory from within, but haven't found a way to do that. Until we find a way we can assert that it's done by the caller.

Merging this PR will:
- Add a `checkHermesOverride` to the host package Gradle build script to ensure `REACT_NATIVE_OVERRIDE_HERMES_DIR` is set (fixing https://github.com/callstackincubator/react-native-node-api/issues/163).
- Drive-by add a test to verify the `linkNodeApiModules` task is calling into the CLI to execute auto-linking.
- Drive-by fix of calling into the CLI to link on Windows.

### DIFF
--- a/.changeset/many-candies-retire.md
+++ b/.changeset/many-candies-retire.md
@@ -1,0 +1,5 @@
+---
+"react-native-node-api": patch
+---
+
+Fix auto-linking from Gradle builds on Windows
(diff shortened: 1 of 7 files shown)


kraenhansen:
Drafting this again, since it seems Gradle is having a hard time executing the CLI on Windows.

PR #243, opened 2025-09-19
Rename target to triplet

kraenhansen:
I renamed `triplet` to `target` in #189, mainly to converge with the naming used in Ferric (inheriting `--target` from the Rust compiler).

I'm backtracking on that decision, because in the context of CMake, developers expect to be able to pass a `--target` to declare the CMake target they want to build.

Merging this PR will:
- Rename "target" back to "triplet" (effectively reverting part of #189).


kraenhansen:
Merging to keep the PR stacking manageable, please review in retrospect 🙏

PR #254, opened 2025-09-24
Use CMake file API to read shared library target paths

### LINKED ISSUE #201: `cmake-rn` fails if projects emit more than one dynamic library
I made the assumption earlier that cmake projects wouldn't emit more than a single dynamic library:

https://github.com/callstackincubator/react-native-node-api/blob/a7cc35ac45e433671fa394b3fec6b44c023f0604/packages/cmake-rn/src/platforms/android.ts#L146

This won't be the case if we want to run the Node.js tests where the `gyp-to-cmake` will produce CMakeLists.txt files declaring multiple output files with different names:

```
cmake_minimum_required(VERSION 3.15)
project(tests-node-api-test_buffer)

add_library(test_buffer SHARED test_buffer.c ${CMAKE_JS_SRC})
set_target_properties(test_buffer PROPERTIES PREFIX "" SUFFIX ".node")
target_include_directories(test_buffer PRIVATE  ${CMAKE_JS_INC})
target_link_libraries(test_buffer PRIVATE ${CMAKE_JS_LIB})
target_compile_features(test_buffer PRIVATE cxx_std_17)
target_compile_definitions(test_buffer PRIVATE NAPI_VERSION=10)

add_library(test_finalizer SHARED test_finalizer.c ${CMAKE_JS_SRC})
set_target_properties(test_finalizer PROPERTIES PREFIX "" SUFFIX ".node")
target_include_directories(test_finalizer PRIVATE  ${CMAKE_JS_INC})
target_link_libraries(test_finalizer PRIVATE ${CMAKE_JS_LIB})
target_compile_features(test_finalizer PRIVATE cxx_std_17)

if(MSVC AND CMAKE_JS_NODELIB_DEF AND CMAKE_JS_NODELIB_TARGET)
  # Generate node.lib
  execute_process(COMMAND ${CMAKE_AR} /def:${CMAKE_JS_NODELIB_DEF} /out:${CMAKE_JS_NODELIB_TARGET} ${CMAKE_STATIC_LINKER_FLAGS})
endif()
```

```
❯ node --run build-tests
Running "cmake-rn" in tests/js-native-api/2_function_arguments to build for React Native
Running "cmake-rn" in tests/node-api/test_buffer to build for React Native
ℹ Using default targets (aarch64-linux-android, arm64-apple-ios-sim)
- Configuring projects (android: aarch64-linux-android / apple: arm64-apple-ios-sim)
✔ Configured projects (android: aarch64-linux-android / apple: arm64-apple-ios-sim)
- Building projects
✔ Built projects
node:internal/process/promises:394
    triggerUncaughtException(err, true /* fromPromise */);
    ^

AssertionError [ERR_ASSERTION]: Expected exactly one library file

2 !== 1

    at file:///Users/kraen.hansen/Repositories/react-native-node-api/packages/cmake-rn/dist/platforms/android.js:104:20
    at async Promise.all (index 0)
    at async Object.postBuild (file:///Users/kraen.hansen/Repositories/react-native-node-api/packages/cmake-rn/dist/platforms/android.js:94:57)
    at async Command.<anonymous> (file:///Users/kraen.hansen/Repositories/react-native-node-api/packages/cmake-rn/dist/cli.js:125:13) {
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: 2,
  expected: 1,
  operator: 'strictEqual'
}
```

### LINKED ISSUE #253: Don't make assumptions on the location of library outputs
`cmake-rn` is making assumptions on the target `LIBRARY_OUTPUT_DIRECTORY` to locate the dynamic libraries to find them and assemble the prebuilds. We [do pass `CMAKE_LIBRARY_OUTPUT_DIRECTORY`](https://github.com/callstackincubator/react-native-node-api/blob/58d17e6cbfa63f6118482cc9df89eac4f355e92a/packages/cmake-rn/src/cli.ts#L327) but this only serves as a default, which the individual target can override.

It seems the most reliable way to get the path of the binary would be to use the [CMake file API](https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html).

kraenhansen:
This PR is stacked on #257 and fixes #253 by introducing use of the [CMake file API](https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html) to read shared library target paths.

Merging this PR will:
- Implement at wrapper around the CMake file API to query and read metadata for targets in the build directories, instead of relying on pre-defined locations.
- Fix #201 by leaning on the CMake structure to group shared library artifacts.

---

> [!NOTE]
> Switches artifact discovery to the CMake File API and updates Android/Apple post-build flows and host helpers to assemble outputs per-target using the new APIs.
> 
> - **cmake-rn**:
>   - Use `cmake-file-api`: add dependency and initialize query (`codemodel` v2) in `src/cli.ts`.
>   - Android `postBuild`: read targets via CMake File API, group shared libraries by target, and assemble per-library `.android.node` outputs using updated host API; include `configuration`.
>   - Apple `postBuild`: read targets via CMake File API, generate frameworks and assemble per-library XCFrameworks (or `.apple.node`).
> - **host (prebuild helpers)**:
>   - Android: change `createAndroidLibsDirectory` signature to `libraries: { triplet, libraryPath }[]`; adjust implementation.
>   - Apple: make `createAppleFramework` async and use `spawn`/`fs.promises`.
> - **ferric**:
>   - Adapt to new Android helper signature and await async `createAppleFramework` when building XCFrameworks.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4d0b7dfbd68a454453b542f955e5e39289695639. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.changeset/evil-vans-love.md
+++ b/.changeset/evil-vans-love.md
@@ -1,0 +1,5 @@
+---
+"cmake-rn": patch
+---
+
+Use CMake file API to read shared library target paths
--- a/package-lock.json
+++ b/package-lock.json
@@ -14808,6 +14808,7 @@
       "version": "0.4.0",
       "dependencies": {
         "@react-native-node-api/cli-utils": "0.1.0",
+        "cmake-file-api": "0.1.0",
         "react-native-node-api": "0.5.1"
       },
       "bin": {
--- a/packages/cmake-rn/package.json
+++ b/packages/cmake-rn/package.json
@@ -23,6 +23,7 @@
   },
   "dependencies": {
     "@react-native-node-api/cli-utils": "0.1.0",
+    "cmake-file-api": "0.1.0",
     "react-native-node-api": "0.5.1"
   },
   "peerDependencies": {
--- a/packages/cmake-rn/src/cli.ts
+++ b/packages/cmake-rn/src/cli.ts
@@ -13,6 +13,7 @@ import {
   wrapAction,
 } from "@react-native-node-api/cli-utils";
 import { isSupportedTriplet } from "react-native-node-api";
+import * as cmakeFileApi from "cmake-file-api";
 
 import {
   getCmakeJSVariables,
@@ -326,6 +327,8 @@ async function configureProject<T extends string>(
     ...options.define,
     { CMAKE_LIBRARY_OUTPUT_DIRECTORY: outputPath },
   ];
+
+  await cmakeFileApi.createSharedStatelessQuery(buildPath, "codemodel", "2");
 
   await spawn(
     "cmake",
--- a/packages/cmake-rn/src/platforms/android.ts
+++ b/packages/cmake-rn/src/platforms/android.ts
@@ -122,7 +122,10 @@ export const platform: Platform<Triplet[], AndroidOpts> = {
     return typeof ANDROID_HOME === "string" && fs.existsSync(ANDROID_HOME);
   },
   async postBuild({ outputPath, triplets }, { autoLink, configuration }) {
-    const prebuilds: Record<string, Partial<Record<Triplet, string>>> = {};
+    const prebuilds: Record<
+      string,
+      { triplet: Triplet; libraryPath: string }[]
+    > = {};
 
     for (const { triplet, buildPath } of triplets) {
       assert(fs.existsSync(buildPath), `Expected a directory at ${buildPath}`);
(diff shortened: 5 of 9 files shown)


matthargett:
Can you cover a few more exceptional cases to make sure there's feedback up to the CLI with actionable output? A couple of cases
* Write an index-v2.json with an unknown version string → assert unsupported-version error
* Drop in a malformed JSON (missing `artifacts` or `targetName`) in either index or reply → assert a JSON-parse exception.
* a `buildDir` with unreadable permissions (common with sloppy `sudo` usage)

I'm okay to save these for a separate PR but before tagging a release. I'm mostly thinking about support load :)

kraenhansen:
I've pushed a commit removing a lot of the schema code I already added.

Instead, I'm going to use CoPilot to generate a wrapper around the CMake file API from their docs, as a separate package, similar to https://crates.io/crates/cmake-file-api for Rust.

kraenhansen:
I've opened #257 now and rebased this PR on top of that.

Issue #270, opened 2025-10-19
Support vendoring Hermes for out-of-tree platforms


shirakaba:
Currently, we look up the appropriate version of `hermes` from `react-native` like so:

https://github.com/callstackincubator/react-native-node-api/blob/04196788819c48840512ab15919e2cac8a509d8a/packages/host/src/node/cli/hermes.ts#L31-L38

For out-of-tree platforms like `react-native-macos`, though, I had to apply the following patch before it'd resolve the correct root during `pod install`:

```js
let reactNativePath;
if(from.endsWith("ios") || from.endsWith("android")){
  reactNativePath = node_path_1.default.dirname(require.resolve("react-native/package.json", {
      // Ensures we'll be patching the React Native package actually used by the app
      paths: [appPackageRoot],
  }));
} else if(from.endsWith("macos")){
  reactNativePath = node_path_1.default.dirname(require.resolve("react-native-macos/package.json", {
      // Ensures we'll be patching the React Native package actually used by the app
      paths: [appPackageRoot],
  }));
} else if(from.endsWith("windows")){
  reactNativePath = node_path_1.default.dirname(require.resolve("react-native-windows/package.json", {
      // Ensures we'll be patching the React Native package actually used by the app
      paths: [appPackageRoot],
  }));
} else {
  throw new Error("Unsupported platform.");
}
```

There is surely a more elegant way to do it (it would be nicest to know the platform for certain, rather than just guessing based on the presence of an `ios`/`android`/`macos` folder in the path), but just to illustrate the point.

kraenhansen:
Another issue here is that we download the Hermes fork into the host package (`node_modules/react-native-node-api/hermes`) and assume there will be a single version needed. Perhaps we should download it into the `react-native` package (or `react-native-macos` / `react-native-windows`) instead? 🤔

kraenhansen:
Reopening the issue, since we'd like to detect the relevant `--react-native-package`, either in [scripts/patch-hermes.rb](https://github.com/callstackincubator/react-native-node-api/blob/main/packages/host/scripts/patch-hermes.rb) or as part of computing a default value inside the CLI.

PR #287, opened 2025-10-23
Add a `--react-native-package` option to `vendor-hermes` command

kraenhansen:
A first attempt at solving #270.

Merging this PR will:
- Add a `--react-native-package` option to the `vendor-hermes` command, controlling what package name to use when determining the base version of Hermes to use for cloning.
- Clone Hermes into `sdks/node-api-hermes` of the React Native package (choosen via `--react-native-package` ☝️)
- Refactor the `scripts/patch-hermes.rb` to allow supplying a `REACT_NATIVE_OVERRIDE_HERMES_DIR` from the outside to skip calling `vendor-hermes` internally.

With this, you should be able to clone and override Hermes in the MacOS out-of-tree package of React Native running (in `<your-app>/ios`):

```bash
REACT_NATIVE_OVERRIDE_HERMES_DIR=$(npx react-native-node-api vendor-hermes --silent --react-native-package react-native-macos) pod install
```

### DIFF
--- a/packages/host/scripts/patch-hermes.rb
+++ b/packages/host/scripts/patch-hermes.rb
@@ -4,13 +4,20 @@ if ENV['RCT_USE_PREBUILT_RNCORE'] == '1'
   raise "React Native Node-API cannot reliably patch JSI when React Native Core is prebuilt."
 end
 
-VENDORED_HERMES_DIR ||= `npx react-native-node-api vendor-hermes --silent '#{Pod::Config.instance.installation_root}'`.strip
-if Dir.exist?(VENDORED_HERMES_DIR)
-  Pod::UI.info "Hermes vendored into #{VENDORED_HERMES_DIR.inspect}"
-else
-  raise "Hermes patching failed. Please check the output above for errors."
+if ENV['REACT_NATIVE_OVERRIDE_HERMES_DIR'].nil?
+  VENDORED_HERMES_DIR ||= `npx react-native-node-api vendor-hermes --silent '#{Pod::Config.instance.installation_root}'`.strip
+  # Signal the patched Hermes to React Native
+  ENV['BUILD_FROM_SOURCE'] = 'true'
+  ENV['REACT_NATIVE_OVERRIDE_HERMES_DIR'] = VENDORED_HERMES_DIR
+elsif Dir.exist?(ENV['REACT_NATIVE_OVERRIDE_HERMES_DIR'])
+  # Setting an override path implies building from source
+  ENV['BUILD_FROM_SOURCE'] = 'true'
 end
 
-# Signal the patched Hermes to React Native
-ENV['BUILD_FROM_SOURCE'] = 'true'
-ENV['REACT_NATIVE_OVERRIDE_HERMES_DIR'] = VENDORED_HERMES_DIR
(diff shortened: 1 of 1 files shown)


packages/host/scripts/patch-hermes.rb
```diff
@@ -4,13 +4,20 @@
   raise "React Native Node-API cannot reliably patch JSI when React Native Core is prebuilt."
 end
 
-VENDORED_HERMES_DIR ||= `npx react-native-node-api vendor-hermes --silent '#{Pod::Config.instance.installation_root}'`.strip
-if Dir.exist?(VENDORED_HERMES_DIR)
-  Pod::UI.info "Hermes vendored into #{VENDORED_HERMES_DIR.inspect}"
-else
-  raise "Hermes patching failed. Please check the output above for errors."
+if ENV['REACT_NATIVE_OVERRIDE_HERMES_DIR'].nil?
+  VENDORED_HERMES_DIR ||= `npx react-native-node-api vendor-hermes --silent '#{Pod::Config.instance.installation_root}'`.strip
```
shirakaba: Is there any reason for this to be `||=` rather than `=`? As `VENDORED_HERMES_DIR` is a local variable, I think it can only be unset at this point?

```suggestion
  VENDORED_HERMES_DIR = `npx react-native-node-api vendor-hermes --silent '#{Pod::Config.instance.installation_root}'`.strip
```

shirakaba (commented):
So to my understanding, this would mean that RNM users can obtain a suitable value for `REACT_NATIVE_OVERRIDE_HERMES_DIR` (which they could then set in their CLI, or just write into their `Podfile`) with:

```sh
npx react-native-node-api vendor-hermes --silent --react-native-package react-native-macos
```

_(your example said `--react-native-package react-native` – correct me if I'm wrong to use `--react-native-package react-native-macos`!)_

I think it's certainly a step forward and makes it possible to proceed, so as it stands currently, **it's worth merging already**.

Though in terms of an ideal implementation, my impression of this approach is that it treats `react-native-macos` as a second-class citizen. RNM already has a thousand extra steps to get set up correctly and it would be nice to not have another gotcha along the way for Node-API modules, especially given that this concerns CocoaPods installs, which are very time-consuming.

I wonder if there's any way we can figure out that the script is being called from a `react-native-macos` app's `Podfile`. Like, inside `packages/host/scripts/patch-hermes.rb`, maybe if we're lucky there's some environment variable that will already have been set by the `Podfile` that we could use to automatically determine that we're running in a `react-native-macos` app, and if so, pass `--react-native-package react-native-macos` along to the `vendor-hermes` script.

packages/host/scripts/patch-hermes.rb
kraenhansen: It's my understanding that Ruby scripts aren't caching requires the same way Node.js would.

I had this originally because it prevents re-execution of the command if the .rb file is (transitively) "required" multiple times.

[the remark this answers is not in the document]
kraenhansen: Exactly and I agree it would be ideal if we didn't vendor for `react-native` if this was driven by a MacOS app. Let's make that happen in a follow-up PR, ideally when we actually have a MacOS app in the repo, testing that flow.

PR #288, opened 2025-10-24
Add x86_64 and universal simulator slices

### LINKED ISSUE #192: Review Apple architectures (add x64_64 slices for all simulators?)
# Background

Running `cmake-rn --apple` produces the following files:

```
build
└── Release
    └── sum.apple.node
        ├── ios-arm64
        ├── ios-arm64-simulator
        ├── macos-arm64_x86_64
        ├── tvos-arm64
        ├── tvos-arm64-simulator
        ├── xros-arm64
        ├── xros-arm64-simulator
        ├── Info.plist
        └── react-native-node-api-module
```

You can review the architectures like this:

```sh
lipo -info build/arm64-apple-ios-sim/out/Release/sum.framework/sum
Non-fat file: build/arm64-apple-ios-sim/out/Release/sum.framework/sum is architecture: arm64
```

At the code level, the architectures are determined here: https://github.com/callstackincubator/react-native-node-api/blob/8557768aa5747c6d51ea1d9b0b10aea4d4404272/packages/cmake-rn/src/platforms/apple.ts#L57-L68

# Issue

I noticed that, for all the simulators, we only build arm64. Shouldn't we build the x86_64 slice as well so that Intel Macs can develop for simulator?

kraenhansen:
Fixes #192 by adding x86_64 and universal simulator slices to the host and refactor `cmake-rn` to pick the right defaults: Universal (arm64;x86_64) over arm64.

### DIFF
--- a/packages/cmake-rn/src/cli.ts
+++ b/packages/cmake-rn/src/cli.ts
@@ -219,6 +219,11 @@ program = program.action(
     const tripletContexts = [...triplets].map((triplet) => {
       const platform = findPlatformForTriplet(triplet);
 
+      assert(
+        platform.isSupportedByHost(),
+        `Triplet '${triplet}' cannot be built, as the '${platform.name}' platform is not supported on a '${process.platform}' host.`,
+      );
+
       return {
         triplet,
         platform,


packages/cmake-rn/src/platforms/types.ts
```diff
@@ -47,9 +48,12 @@ export type Platform<
    */
   triplets: Readonly<Triplets>;
   /**
-   * Get the limited subset of triplets that should be built by default for this platform, to support a development workflow.
+   * Get the limited subset of triplets that should be built by default for this platform.
+   *
    */
-  defaultTriplets(): Triplets[number][] | Promise<Triplets[number][]>;
+  defaultTriplets(
+    purpose: "development" | "release",
```
shirakaba: Xcode uses the terminology [build configuration](https://developer.apple.com/documentation/xcode/adding-a-build-configuration-file-to-your-project) (which you can name `"orange" | "banana"` if you wish, but defaults to `Debug` and `Release` when creating new projects).

They confusingly use the term `Development` for codesigning for local deployment.

shirakaba (approved):
Looks like a great improvement, thank you!

packages/cmake-rn/src/platforms/types.ts
shirakaba: Oh right. Then surely this should be changed to `onlyBuildActiveArchitecture: boolean` or something? As "active architecture" I believe is terminology commonly used across various IDEs and build systems.

Or `buildAllArchitectures: boolean`.

kraenhansen: I agree - it's the same concept as the `activeArchOnly` option passable to the React Native CLI: [src/commands/buildAndroid/index.ts#L58-L77](https://github.com/react-native-community/cli/blob/dc848fd90842beb68e92f6ab7c9a00458a151b03/packages/cli-platform-android/src/commands/buildAndroid/index.ts#L58-L77) where it's actually probing ADB. We might be able to do the same or something similar to figure out what simulators and emulators are actively booted 🤔

That being said, I don't know if it's worth it though investing too much into this feature: We're running all of the builds of triplets in parallel so the increase in build-time doesn't scale linearly with the amount of triplets.

PR #329, opened 2025-11-10
Enable multiple Node-API hosting JS engines / runtimes to share the global Node-API functions

kraenhansen:
This is my suggestion for adding "multi-host" support to `weak-node-api`, enabling multiple engines implementing Node-API to co-exist and share the Node-API function namespace. While not needed specifically for bringing Node-API to React Native adding this could make `weak-node-api` more applicable in other scenarios where multiple engines implementing Node-API share a single process.

I'm proposing adding mechanisms to "wrap" the opaque pointers of specific Node-API implementors with references to the `NodeApiHost` object to enable deferring implementation of Node-API functions based on the `napi_env`, `node_api_basic_env`, `napi_threadsafe_function` or `napi_async_cleanup_hook_handle` passed.

```mermaid
classDiagram
    class NodeApiHost {
        <<interface>>
    }

    class NodeApiMultiHost {
        -vector~unique_ptr~WrappedEnv~~ envs
        +wrap(napi_env, weak_ptr~NodeApiHost~) napi_env
        +static napi_* methods...
    }

    class WrappedEnv {
        +napi_env value
        +weak_ptr~NodeApiHost~ host
        -vector~unique_ptr~WrappedThreadsafeFunction~~ threadsafe_functions
        -vector~unique_ptr~WrappedAsyncCleanupHookHandle~~ async_cleanup_hook_handles
        +wrap(napi_threadsafe_function, WrappedEnv*, weak_ptr~NodeApiHost~) napi_threadsafe_function
        +wrap(napi_async_cleanup_hook_handle, WrappedEnv*, weak_ptr~NodeApiHost~) napi_async_cleanup_hook_handle
    }

    class WrappedThreadsafeFunction {
        +napi_threadsafe_function value
        +WrappedEnv* env
        +weak_ptr~NodeApiHost~ host
    }

    class WrappedAsyncCleanupHookHandle {
        +napi_async_cleanup_hook_handle value
        +WrappedEnv* env
        +weak_ptr~NodeApiHost~ host
    }

    NodeApiMultiHost --|> NodeApiHost : inherits
    NodeApiMultiHost *-- "0..*" WrappedEnv : owns
    WrappedEnv *-- "0..*" WrappedThreadsafeFunction : owns
    WrappedEnv *-- "0..*" WrappedAsyncCleanupHookHandle : owns
    WrappedThreadsafeFunction --> WrappedEnv : references
    WrappedAsyncCleanupHookHandle --> WrappedEnv : references
    WrappedEnv --> NodeApiHost : weak reference
    WrappedThreadsafeFunction --> NodeApiHost : weak reference
    WrappedAsyncCleanupHookHandle --> NodeApiHost : weak reference
    WrappedEnv ..|> enable_shared_from_this : implements

    note for NodeApiMultiHost "Manages multiple Node-API host implementations"
    note for WrappedEnv "Wraps napi_env with ownership tracking for threadsafe functions and async cleanup hook handles"
```

`WrappedEnv`, `WrappedThreadsafeFunction` and `WrappedAsyncCleanupHookHandle` objects can then be passed around like their respective opaque pointers and are "unwrapped" in the internal implementation of the "multi host" implementation of Node-API functions.

`Wrapped` objects are created calling one of the `wrap` instance methods on `NodeApiMultiHost` or `WrappedEnv`, which are called internally in `napi_create_threadsafe_function` and `napi_add_async_cleanup_hook` too.

## Usage

- Create a `NodeApiMultiHost` (providing functions to register modules and handling a fatal error)
- Inject it as the global host
- Get an env from the actual Node-API implementatin
- Wrap the env with using the `NodeApiMultiHost` (calling `multi_host.wrap(original_env, host);`)
- Pass the wrapped env to Node-API functions to delegate as needed

https://github.com/callstackincubator/react-native-node-api/blob/f79a30bf7338f1bb1f65e68790f960d3fa620037/packages/weak-node-api/tests/test_multi_host.cpp#L18-L52

## TODO

- [ ] Provide a way to "delete" a `WrappedEnv` (besides deleting the entire `NodeApiMultiHost`)
- [ ] Delete `WrappedThreadsafeFunction` (in some `napi_*_threadsafe_function` function?) and `WrappedAsyncCleanupHookHandle` (in `napi_remove_async_cleanup_hook`).

## Open questions

1. Should we use `std::function` instead of raw function pointers for all (or some) of the `WeakNodeApiHost` members? This would allow capturing lambdas, making it much easier to provide a meaningful implementation of for example `napi_module_register`.

## Generated code

Below are samples from the generated code:

### `napi_create_object` implementation

```cpp
napi_status NodeApiMultiHost::napi_create_object(napi_env arg0,
                                                 napi_value *arg1) {
  auto wrapped = reinterpret_cast<WrappedEnv *>(arg0);
  if (auto host = wrapped->host.lock()) {
    if (host->napi_create_object == nullptr) {
      fprintf(stderr, "Node-API function 'napi_create_object' called on a host "
                      "which doesn't provide an implementation\n");
      return napi_status::napi_generic_failure;
    }

    return host->napi_create_object(wrapped->value, arg1);

  } else {
    fprintf(stderr, "Node-API function 'napi_create_object' called after host "
                    "was destroyed.\n");
    return napi_status::napi_generic_failure;
  }
};
```

### `napi_create_threadsafe_function` and `napi_add_async_cleanup_hook` implementations

Notice the calls to `wrap`, wrapping their opaque "out" pointers.

```cpp
napi_status NodeApiMultiHost::napi_create_threadsafe_function(
    napi_env env, napi_value func, napi_value async_resource,
    napi_value async_resource_name, size_t max_queue_size,
    size_t initial_thread_count, void *thread_finalize_data,
    napi_finalize thread_finalize_cb, void *context,
    napi_threadsafe_function_call_js call_js_cb,
    napi_threadsafe_function *result) {
  auto wrapped = reinterpret_cast<WrappedEnv *>(env);
  if (auto host = wrapped->host.lock()) {
    if (host->napi_create_threadsafe_function == nullptr) {
      fprintf(stderr,
              "Node-API function 'napi_create_threadsafe_function' called on a "
              "host which doesn't provide an implementation\n");
      return napi_status::napi_generic_failure;
    }

    auto status = host->napi_create_threadsafe_function(
        wrapped->value, func, async_resource, async_resource_name,
        max_queue_size, initial_thread_count, thread_finalize_data,
        thread_finalize_cb, context, call_js_cb, result);
    if (status == napi_status::napi_ok) {
      *result = wrapped->wrap(*result, wrapped, wrapped->host);
    }
    return status;

  } else {
    fprintf(stderr, "Node-API function 'napi_create_threadsafe_function' "
                    "called after host was destroyed.\n");
    return napi_status::napi_generic_failure;
  }
};

napi_status NodeApiMultiHost::napi_add_async_cleanup_hook(
    node_api_basic_env env, napi_async_cleanup_hook hook, void *arg,
    napi_async_cleanup_hook_handle *remove_handle) {
  auto wrapped = reinterpret_cast<WrappedEnv *>(env);
  if (auto host = wrapped->host.lock()) {
    if (host->napi_add_async_cleanup_hook == nullptr) {
      fprintf(stderr, "Node-API function 'napi_add_async_cleanup_hook' called "
                      "on a host which doesn't provide an implementation\n");
      return napi_status::napi_generic_failure;
    }

    auto status = host->napi_add_async_cleanup_hook(wrapped->value, hook, arg,
                                                    remove_handle);
    if (status == napi_status::napi_ok) {
      *remove_handle = wrapped->wrap(*remove_handle, wrapped, wrapped->host);
    }
    return status;

  } else {
    fprintf(stderr, "Node-API function 'napi_add_async_cleanup_hook' called "
                    "after host was destroyed.\n");
    return napi_status::napi_generic_failure;
  }
};
```

### DIFF
--- a/packages/weak-node-api/tests/test_inject.cpp
+++ b/packages/weak-node-api/tests/test_inject.cpp
@@ -23,3 +23,42 @@ TEST_CASE("inject_weak_node_api_host") {
     REQUIRE(called);
   }
 }
+
+TEST_CASE("calling into host from functions") {
+  auto my_create_function = [](napi_env arg0, const char *arg1, size_t arg2,
+                               napi_callback arg3, void *arg4,
+                               napi_value *arg5) -> napi_status {
+    // This is a failing noop as we're not actually creating a JS functions
+    return napi_status::napi_generic_failure;
+  };
+  NodeApiHost host{.napi_create_function = my_create_function};
+  inject_weak_node_api_host(host);
+  napi_env raw_env{};
+
+  SECTION("via global function") {
+    napi_value result;
+    napi_callback cb = [](napi_env env, napi_callback_info info) -> napi_value {
+      napi_value obj;
+      napi_status status = napi_create_object(env, &obj);
+      return obj;
+    };
+
+    napi_create_function(raw_env, "foo", 3, cb, nullptr, &result);
+  }
+
+  SECTION("via callback info") {
+    napi_value result;
+    napi_callback cb = [](napi_env env, napi_callback_info info) -> napi_value {
+      // Get host via callback info
+      void *data;
+      napi_get_cb_info(env, info, nullptr, nullptr, nullptr, &data);
+      auto *host_ptr = static_cast<decltype(&host)>(data);
+
+      napi_value obj;
+      napi_status status = host_ptr->napi_create_object(env, &obj);
+      return obj;
+    };
+
+    napi_create_function(raw_env, "foo", 3, cb, &host, &result);
+  }
+}


packages/weak-node-api/scripts/generators/NodeApiMultiHost.ts
```diff
@@ -0,0 +1,263 @@
...
+    
+    #include <node_api.h>
+
+    #include "NodeApiHost.hpp"
+
+    struct NodeApiMultiHost : public NodeApiHost {
+
+      struct WrappedEnv;
+
+      struct WrappedThreadsafeFunction {
```
RobinWuu: I noticed that we're calling the constructors of `struct WrappedThreadsafeFunction` and `struct WrappedAsyncCleanupHookHandle` with three arguments:
```c++
auto ptr = std::make_unique<WrappedAsyncCleanupHookHandle>(original, env, weak_host);
auto ptr = std::make_unique<WrappedThreadsafeFunction>(original, env, weak_host);
```
When compiling locally, the compiler throws an error. I'm wondering if these structs should explicitly define a constructor that accepts these three parameters?😊

```
/Applications/Xcode_15.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.2.sdk/usr/include/c++/v1/__memory/unique_ptr.h:686:30: error: no matching constructor for initialization of 'NodeApiMultiHost::WrappedAsyncCleanupHookHandle'
  return unique_ptr<_Tp>(new _Tp(_VSTD::forward<_Args>(__args)...));
                             ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
weak-node-api/generated/NodeApiMultiHost.cpp:240:12: note: in instantiation of function template specialization 'std::make_unique<NodeApiMultiHost::WrappedAsyncCleanupHookHandle, napi_async_cleanup_hook_handle__ *&, NodeApiMultiHost::WrappedEnv *&, std::weak_ptr<NodeApiHost> &>' requested here
      std::make_unique<WrappedAsyncCleanupHookHandle>(original, env, weak_host);
```

RobinWuu:
Hi, @kraenhansen ,
I've run into an issue: in the callback function of `napi_create_function`, the obtained `napi_env` is not a `WrappedEnv` but the original `napi_env`. This prevents direct use of the Weak Node API within the callback.
Perhaps we should consider adding a mapping from `napi_env` to `WeakNodeApiHost` via a host getter? I can't think of a better approach for now.
Thanks for your work on this! And I look forward to hearing your thoughts on this issue.

```c++
auto host_foo = std::shared_ptr<WeakNodeApiHost>(new WeakNodeApiHost{
    .napi_create_function =
        [](napi_env env, const char* utf8name, size_t length, napi_callback cb,
           void* callback_data, napi_value* result) -> napi_status {
      // Omitted: actual implementation of napi_create_function across different JS engines
      return napi_ok;
    }});

// Create and inject a multi-host, then wrap two envs
WeakNodeApiMultiHost multi_host{nullptr, nullptr};
inject_weak_node_api_host(multi_host);

// Original napi_env
napi_env raw_env{};
// foo_env is a WrappedEnv
auto foo_env = multi_host.wrap(napi_env{}, host_foo);

napi_callback foo_cb = [](napi_env env, napi_callback_info info) -> napi_value {
  // This `env` is not the WrappedEnv foo_env, but the original napi_env raw_env.
  // Cannot directly call weak-node-api functions here.
  return nullptr;
};

napi_value foo_fn = nullptr;
napi_create_function(foo_env, "foo", 3, foo_cb, nullptr, &foo_fn);
```

RobinWuu:
> > Cannot directly call weak-node-api functions here.
> 
> Ideally, you'd be able to capture the `host_foo` as a `weak_ptr` when creating `foo_cb`, but this isn't possible as `napi_create_function` expects raw function pointers. I've added two tests showing how I imagine this could be done:
> 
> https://github.com/callstackincubator/react-native-node-api/blob/e82efc209a032738c09c76426d27687c5a11277d/packages/weak-node-api/tests/test_inject.cpp#L38-L63
> 
> Perhaps you could adjust these or add a new test as an example of the issues you're seeing?

I've added a test case try to illustrate my issues. 🙏   
https://github.com/RobinWuu/react-native-node-api/blob/5b4eb3f91fcc8e0e9856b8006952e67cb7def7bb/packages/weak-node-api/tests/test_multi_host.cpp#L174-L206    

By the way, I've experimented with an alternative approach locally: injecting a `NodeApiHostGetter` to handle the mapping from `napi_env` (which can be `nullptr`, e.g., in `napi_call_threadsafe_function`) to `NodeApiHost`.

This approach solves the issues I'm facing because it avoids having two `napi_env`. However, it intrudes upon the weak-node-api in normal mode, meaning that even in non-multi-host scenarios, one must indirectly obtain the `NodeApiHost` through the `NodeApiHostGetter`.

This is the commit for `NodeApiHostGetter` injection approach, in my fork repo: https://github.com/RobinWuu/react-native-node-api/commit/f233545ecfc2aab3ac41d3650c3a60cf9f891e86

And here's the related test case: https://github.com/RobinWuu/react-native-node-api/commit/8c68ff4e1306967686ca75f6f34269abaa294297

kraenhansen:
Okay, reading the comments in depth, I believe I understand your need and problem 👍 Thanks for bearing with me 🙂

This line in my test wont work in a real scenario, since the global functions expect a wrapped env and I'm calling it with a raw unwrapped env: https://github.com/callstackincubator/react-native-node-api/blob/e82efc209a032738c09c76426d27687c5a11277d/packages/weak-node-api/tests/test_inject.cpp#L42

RobinWuu:
Yes, passing the `node_api_basic_env` to a `NodeApiHostGetter` works in my local environment. Following this approach, I've made local adaptations to integrate the NAPI implementations of Lynx's three JS engines (V8, JSC, QuickJS) into `weak-node-api`.

The implementation idea of `NodeApiHostGetter` is: if the passed `node_api_basic_env` is a null pointer, return a default global host. If the passed `node_api_basic_env` is not a null pointer, find and return the host corresponding to the env through a mapping table.

```c++
    static NodeApiHost sDefaultHost{.napi_call_threadsafe_function = my_call_threadsafe_function,
                                    .napi_create_object = dummy_create_object};
    static std::unordered_map<uint64_t, std::shared_ptr<NodeApiHost>> sMultiHost;
    inject_weak_node_api_host_getter([](node_api_basic_env env) -> NodeApiHost* {
      if (env == nullptr) {
        return &sDefaultHost;
      }
      return sMultiHost[reinterpret_cast<uint64_t>(env)].get();
    });
```

This relies on some modifications to `weak-node-api`: if the current Node-API has no `napi_env`/`node_api_basic_env` parameter, pass `nullptr` to `NodeApiHostGetter`. Here are the changes I attempted in my fork repository: https://github.com/RobinWuu/react-native-node-api/commit/f233545ecfc2aab3ac41d3650c3a60cf9f891e86

On this basis, I tried writing a unit test to further illustrate the implementation of `NodeApiHostGetter`:
https://github.com/RobinWuu/react-native-node-api/blob/p/robin/napi-host-getter/packages/weak-node-api/tests/test_inject.cpp#L13

PR #342, opened 2026-01-27
React Native 0.83.1 (0.83.0, 0.82.1 & 0.82.0)

kraenhansen:
Merging this PR will:
- Bump the test app's React Native version and declare peer dependency support up to v0.83.1.
- Fix escaping bundle identifiers in `gyp-to-cmake` (I have no idea why that worked before 🙈 )


kraenhansen:
Seems to outdated to be valuable with the current state of the `next` branch targeting React Native `0.88.0-nightly-20260809-db662caea`

PR #351, opened 2026-02-10
Apple link without cocoapods

### LINKED ISSUE #23: Namespace vendored Xcframeworks per app
At the time of writing, we have only one test app, but in a mono-repo with more than one (which I imagine we'll get at some point - with individual examples, etc) our host package will be sharing the `./xcframework` directory across multiple apps. Running "pod install" in one followed by "pod install" in another will result in deletions of the previous Xcframeworks on the second run, which is impractical. If we instead found a way to store the vendored Xcframeworks in the app's `ios/Pods` directory, we'd be able to avoid this "overriding" behaviour when switching between two (or more) apps.

### LINKED ISSUE #213: Script phase to auto-link executes too late
When rebuilding the app, changes to add-ons don't materialize on the first, but instead the second rebuild of the app. I believe this is because the auto-linking script phase executes too late. We should set `:execution_position => :before_compile` to make it execute sooner in the build process.

kraenhansen:
Fixes #23, #213 and #262 by modifying the Xcode project after it has been generated by CocoaPods.

Merging this PR will:
- Add a `patch-xcode-project` command to the host CLI: When called, it will modify the main Xcode app target[^1] to include a build phase which calls into the CLI with `link --apple` args.
- Add an exit hook to the host package's podspec to conveniently run the `patch-xcode-project` command. Since users can run this commend themselves or we could drive it from a SPM project, I don't consider this a dependency on CocoaPods as such.
- Modify the Apple link command to expect invocation from inside a build phase, to copy, renaming and signing frameworks (not xcframeworks) to their final destination in the app 
bundle, based on environment variables provided by Xcode.

Because the frameworks are no longer explicitly referenced by targets we also don't have the (currently untracked) issue of the dynamic libraries being loaded at startup time. Which means faster boot times and less concern about the order of operations as they'll be loaded when required by the app.

[^1]: We could have added it as a build phase to our own host package's pod target, but then we'd need an additional phase to move the frameworks slices from the pod target into the app bundle - which seems wasteful. Instead I suggest just driving the discovery and embedding of addon frameworks directly from a build phase on the app target.

### DIFF
--- a/packages/host/react-native-node-api.podspec
+++ b/packages/host/react-native-node-api.podspec
@@ -3,19 +3,7 @@ require "json"
 package = JSON.parse(File.read(File.join(__dir__, "package.json")))
 
 require_relative "./scripts/patch-hermes"
-
-NODE_PATH ||= `which node`.strip
-CLI_COMMAND ||= "'#{NODE_PATH}' '#{File.join(__dir__, "dist/node/cli/run.js")}'"
-COPY_FRAMEWORKS_COMMAND ||= "#{CLI_COMMAND} link --apple '#{Pod::Config.instance.installation_root}'"
-
-# We need to run this now to ensure the xcframeworks are copied vendored_frameworks are considered
-XCFRAMEWORKS_DIR ||= File.join(__dir__, "xcframeworks")
-unless defined?(@xcframeworks_copied)
-  puts "Executing #{COPY_FRAMEWORKS_COMMAND}"
-  system(COPY_FRAMEWORKS_COMMAND) or raise "Failed to copy xcframeworks"
-  # Setting a flag to avoid running this command on every require
-  @xcframeworks_copied = true
-end
+require_relative "./scripts/patch-xcode-project"
 
 if ENV['RCT_NEW_ARCH_ENABLED'] == '0'
   Pod::UI.warn "React Native Node-API doesn't support the legacy architecture (but RCT_NEW_ARCH_ENABLED == '0')"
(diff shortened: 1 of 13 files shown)


kraenhansen:
That's weird 🙄 Running the app through the CLI instead of XCode doesn't seem to fail validating the app 🤷‍♂️

packages/host/react-native-node-api.podspec
```diff
@@ -35,16 +23,6 @@ Pod::Spec.new do |s|
...
 
-  s.vendored_frameworks = "auto-linked/apple/*.xcframework"
-  s.script_phase = {
-    :name => 'Copy Node-API xcframeworks',
-    :execution_position => :before_compile,
-    :script => <<-CMD
-      set -e
-      #{COPY_FRAMEWORKS_COMMAND}
-    CMD
-  }
```
kraenhansen: This is one of the most important changes of the PR - we no longer have to reference the addon xcframeworks from our podspec.

packages/host/types/xmldom.d.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * Local type declaration for @xmldom/xmldom that does not pull in lib "dom".
```
shirakaba: Nice 👍

shirakaba (approved):
All changes look like good upgrades! Some minor suggestions only.

packages/host/src/node/cli/apple.ts
```diff
@@ -3,16 +3,77 @@ import path from "node:path";
...
+  assert(
+    fs.existsSync(pbxprojPath),
+    `Expected a project.pbxproj file at '${pbxprojPath}'`,
+  );
+  const xcodeProject = xcode.XcodeProject.open(pbxprojPath);
+  // Create a build phase in all application targets to stage and rename the addon frameworks
+  const applicationTargets = xcodeProject.rootObject.props.targets.filter(
+    (target) =>
+      xcode.PBXNativeTarget.is(target) &&
+      target.props.productType === "com.apple.product-type.application",
```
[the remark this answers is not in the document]
kraenhansen: It is passing on both iOS and MacOS 👍 This is also the check Evan uses internally in https://github.com/EvanBacon/xcode/blob/a71555f5c0c40afb3a450c7074988823d2a8d607/src/api/PBXProject.ts#L195C11-L200

Looking at [the valid values for this](https://github.com/EvanBacon/xcode/blob/a71555f5c0c40afb3a450c7074988823d2a8d607/src/json/types.ts#L66-L90), we could probably extend this in the future to support:
- `"com.apple.product-type.application.on-demand-install-capable"`
- `"com.apple.product-type.application.watchapp"`
- `"com.apple.product-type.application.watchapp2"`
- `"com.apple.product-type.application.watchapp2-container"`
- `"com.apple.product-type.application.messages"`

PR #356, opened 2026-02-12
Fix `watchFolders` when building MacOS app on CI

kraenhansen:
This fixes `watchFolders` when building MacOS app on CI, a new version of rnx-kit started including a watchFolder which disabled the check we did to inject the package root as watch folder.

Merging this PR will:
- Update the metro config (copied from the iOS app) to use the project root to determine when to inject the package root as watch folder.
- Start bundling and building the MacOS app in separate steps to get better feedback and an environment closer what we want to ensure stability in.

### DIFF
--- a/apps/test-app/metro.config.js
+++ b/apps/test-app/metro.config.js
@@ -11,9 +11,7 @@ const config = makeMetroConfig({
   },
 });
 
-console.log("watchFolders =", config.watchFolders);
-
-if (config.watchFolders.length === 0) {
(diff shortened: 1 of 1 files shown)


kraenhansen:
@cursor help me understand what is different from this run (failing) vs https://github.com/callstackincubator/react-native-node-api/pull/352

I suspect the MacOS runner or Xcode got updated.

kraenhansen:
@cursor help debug this failure on CI.

PR #360, opened 2026-02-14
Bundle, build and run iOS test app in separate steps in production + ccache enabled

kraenhansen:
Getting some inspiration from [prior art](https://github.com/realm/realm-js/blob/main/.github/workflows/pr-realm-js.yml).


kraenhansen:
Superseded by #427

Issue #366, opened 2026-02-27
Cannot import node-api module


owl352:
Good day! I started preparing my module for release and encountered a problem during testing in React Native: I couldn't import the napi module either through require with `babel.config.js` or through `requireNodeAddon`. 

Error:
```
[DEBUG] [NodeApiHost] [pshenmic-dpp--pshenmic_dpp] Loading addon by '@rpath/pshenmic-dpp--pshenmic_dpp.framework/pshenmic-dpp--pshenmic_dpp'
[...]

Issue #54, opened 2025-05-14
Scan transitive dependencies when linking Node-API modules


kraenhansen:
Transitive dependencies (i.e. a dependency of a dependency of the app) aren't crawled when auto-linking, which means a library must be a direct dependency of an app to be considered when auto-linking Node-API addons. We're currently limiting the scanning for dependencies to direct dependencies of the app.

This is [a known limitation of the React Native community CLI](https://shift.infinite.red/transitive-dependencies-and-react-native-autolinking-d4d4b7a651cf) too and a design choice.
It would however be great if we could supported scanning for Node-API in dependencies of dependencies, etc: Having this would make it simpler for libraries to ship with dependencies on Node-API modules, since it could declare its dependency directly instead of using a peer dependency and asking the developer to declare the dependency on its behalf.

As a preliminary workaround, we could specify a `package.json` config object read when auto-linking a direct dependency of an app, which could enumerate transitive dependencies that needed to be scanned as well.

PR #367, opened 2026-02-28
Add `reactNativeNodeApi ` field support for `package.json`

owl352:
# Issue
At the moment, we cannot import Node-API dependencies unless they are explicitly declared in package.json #366. However, there are valid use cases where a user installs a package that depends on an Node-API module. For example, it could be an SDK that includes the Node-API module.

Right now, there is only 2 workaround: manually adding that package to dependencies or manually linking. This is far from ideal, as it complicates the migration flow between versions of the dependency that relies on this package.
I see only two potential solutions at the moment: scan the entire `node_modules` directory, or get a “hint” from somewhere about which modules require auto-linking. I consider the first option to be bad, as it requires a lot of resources.

The goal of this PR is to introduce a separate configuration field in `package.json` that allows specifying package names for auto-linking, without interacting with the package manager in any way.

# Things done
- Updated `package-lock.json`
- Added `reactNativeNodeApi` field reading from `package.json` for project and depencies for project

new `package.json` example

```json
{
  "name": "expo-pshenmic-dpp-test",
  "version": "1.0.0",
  "main": "index.ts",
  "scripts": {
    "start": "expo start",
  },
  "dependencies": {
    "moduleA": "^54.0.6",
  },
  "reactNativeNodeApi": {
    "scan": {
      "dependencies": ["moduleA"]
    }
  },
  "devDependencies": {},
  "private": true
}
```

The module maintainers can also set the configuration within the module's package.json file. In this case, the configuration section need not be duplicated in the project's package.json.

### DIFF
--- a/packages/host/src/node/path-utils.test.ts
+++ b/packages/host/src/node/path-utils.test.ts
@@ -541,6 +541,50 @@ describe("findNodeApiModulePathsByDependency", () => {
     });
   });
 
+  it("should find Node-API paths by dependency in root package.json with bad configuration (excluding certain packages)", async (context) => {
+    const packagesNames = ["lib-a", "lib-b", "lib-c", "lib-d", "lib-e"];
+    const tempDir = setupTempDirectory(context, {
+      "app/package.json": JSON.stringify({
+        name: "app",
+        dependencies: Object.fromEntries(
+          packagesNames
+            .slice(0, 2)
+            .map((packageName) => [packageName, "^1.0.0"]),
+        ),
+        reactNativeNodeApi: {
+          scan: "lib-e",
+        },
+      }),
+      ...Object.fromEntries(
+        packagesNames.map((packageName) => [
+          `app/node_modules/${packageName}`,
+          {
+            "package.json": JSON.stringify({
+              name: packageName,
+              main: "index.js",
+            }),
+            "index.js": "",
+            "addon.apple.node/react-native-node-api-module": "",
+          },
+        ]),
+      ),
+    });
+
+    // shouldn't drop error
+    const result = await findNodeApiModulePathsByDependency({
+      fromPath: path.join(tempDir, "app"),
+      platform: "apple",
+      includeSelf: false,
+      excludePackages: ["lib-a"],
+    });
+    assert.deepEqual(result, {
+      "lib-b": {
+        path: path.join(tempDir, "app/node_modules/lib-b"),
+        modulePaths: ["addon.apple.node"],
+      },
+    });
+  });
+
   it("should find Node-API paths by dependency in root package.json configuration with incorrect dependency configuration (excluding certain packages)", async (context) => {
     const packagesNames = ["lib-a", "lib-b", "lib-c", "lib-d"];
     const tempDir = setupTempDirectory(context, {
@@ -706,7 +750,7 @@ describe("findNodeApiModulePathsByDependency", () => {
         packagesNames.slice(1).map((packageName, i) => {
           // if even then i-1
           // if not even then i+1
-          const dependencyIndex = i + ((i % 2) * 2 - 1)
+          const dependencyIndex = i + ((i % 2) * 2 - 1);
 
           return [
             `app/node_modules/${packageName}`,
--- a/packages/host/src/node/path-utils.test.ts
+++ b/packages/host/src/node/path-utils.test.ts
@@ -657,25 +657,31 @@ describe("findNodeApiModulePathsByDependency", () => {
         ),
       }),
       ...Object.fromEntries(
-        packagesNames.slice(1).map((packageName, i) => [
-          `app/node_modules/${packageName}`,
-          {
-            "package.json": JSON.stringify({
-              name: packageName,
-              main: "index.js",
-              reactNativeNodeApi: {
-                scan: {
-                  dependencies:
-                    packagesNames[i + ((i % 2) * 2 - 1)] != null
-                      ? [packagesNames[i + ((i % 2) * 2 - 1)]]
-                      : [],
+        packagesNames.slice(1).map((packageName, i) => {
+          // if even then i-1
+          // if not even then i+1
+          const dependencyIndex = i + ((i % 2) * 2 - 1)
+
+          return [
+            `app/node_modules/${packageName}`,
+            {
+              "package.json": JSON.stringify({
+                name: packageName,
+                main: "index.js",
+                reactNativeNodeApi: {
+                  scan: {
+                    dependencies:
+                      packagesNames[dependencyIndex] != null
+                        ? [packagesNames[dependencyIndex]]
+                        : [],
+                  },
                 },
-              },
-            }),
-            "index.js": "",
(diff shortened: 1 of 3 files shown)


kraenhansen:
Overall - I agree this is a problem worth solving 👍 and it seems related to #54.

As a first, we're consciously using "Node-API" / `nodeApi` instead of "N-API" / `napi` as this is [the name the Node.js project choose going forward](https://openjsf.org/blog/node-js-update-renaming-n-api-to-node-api) and the name used in this proposal is not aligned with that.

kraenhansen:
We've actively worked towards putting as few restrictions on the packages bringing Node-API modules as possible, to make it more likely to migrate / re-use existing packages intended for Node.js.

I do however see value in (optional) package-specific configuration intended for our tooling with linking efficiently - wrote down a few thoughts that I'd love your input on https://github.com/callstackincubator/react-native-node-api/issues/368

packages/host/src/node/path-utils.ts
```diff
@@ -294,41 +294,73 @@ export function visualizeLibraryMap(libraryMap: LibraryMap) {
...
+  return { reactNativeNodeApi };
+}
+
+export interface ReactNativeNodeAPIConfiguration {
+  reactNativeNodeApi?: {
+    scan?: {
+      dependencies?: string[];
+    };
+  };
+}
```
kraenhansen: I'd probably use [zod](https://npmjs.org/zod) here (already a dependency of the host package), to get runtime errors if the value in the package doesn't match what we expect.

owl352: Done ✅

kraenhansen:
I would have probably assumed this would be added in the `findNodeApiModulePathsByDependency` implementation 🤔 Perhaps now is a good time to add a couple of tests to make sure we're on the same page in regards to the semantics of the implementation.

One thing I'm wondering is how deep this implementation goes. We want to ensure we support transitive dependencies of arbitrary depth and we probably also want to ensure we break on cycles.

owl352:
@kraenhansen Okay, I'll add the tests a little later. At the moment, there are actually only two layers: we read the root `package.json`, as well as the `package.json` dependencies from the the root `package.json`

owl352:
> I would have probably assumed this would be added in the `findNodeApiModulePathsByDependency` implementation 🤔 Perhaps now is a good time to add a couple of tests to make sure we're on the same page in regards to the semantics of the implementation.
> 
> One thing I'm wondering is how deep this implementation goes. We want to ensure we support transitive dependencies of arbitrary depth and we probably also want to ensure we break on cycles.

Added recursive configuration processing, and depth is now limited only by configurations in `package.json`.  Added a unit tests.

packages/host/src/node/path-utils.test.ts
```diff
@@ -491,6 +491,318 @@ describe("findNodeApiModulePathsByDependency", () => {
...
+          {
+            "package.json": JSON.stringify({
+              name: packageName,
+              main: "index.js",
+              reactNativeNodeApi: {
+                scan: {
+                  dependencies:
+                    packagesNames[i + ((i % 2) * 2 - 1)] != null
+                      ? [packagesNames[i + ((i % 2) * 2 - 1)]]
+                      : [],
```
kraenhansen: 🤔 What's the idea here?

packages/host/src/node/path-utils.ts
```diff
@@ -294,33 +295,104 @@ export function visualizeLibraryMap(libraryMap: LibraryMap) {
...
+    visited.add(name);
+
+    const root = resolvePackageRoot(requireFromRoot, name);
+    if (!root) {
+      continue;
+    }
+
+    result[name] = root;
+
+    const config = findPackageConfigurationByPath(root);
```
kraenhansen, line 400: I think the processing of each package should be try-catch'ed (printing any error as warning) to ensure one malformed package doesn't break everything.

packages/host/src/node/path-utils.test.ts
```diff
@@ -491,6 +491,318 @@ describe("findNodeApiModulePathsByDependency", () => {
   });
 });
 
+describe("findNodeApiModulePathsByConfiguration", () => {
```
kraenhansen:
```suggestion
describe("findNodeApiModulePathsByDependency", () => {
```

packages/host/src/node/path-utils.ts
kraenhansen: It would also be great with a test to verify how it behaves on malformed packages.

packages/host/src/node/path-utils.ts
```diff
@@ -294,33 +295,104 @@ export function visualizeLibraryMap(libraryMap: LibraryMap) {
...
+
+    if (visited.has(name)) {
+      continue;
+    }
+    visited.add(name);
+
+    const root = resolvePackageRoot(requireFromRoot, name);
+    if (!root) {
+      continue;
+    }
```
kraenhansen, line 396: It would be great emit a warning here and have a test to ensure this behavior.

packages/host/src/node/path-utils.test.ts
```diff
@@ -491,6 +491,318 @@ describe("findNodeApiModulePathsByDependency", () => {
...
+          },
+        ]),
+      ),
+    });
+
+    const result = await findNodeApiModulePathsByDependency({
+      fromPath: path.join(tempDir, "app"),
+      platform: "apple",
+      includeSelf: false,
+      excludePackages: ["lib-a"],
```
kraenhansen: Is this actively used in the test?

packages/host/src/node/path-utils.ts
owl352: it is impossible to form it incorrectly. Each field of the configuration scheme, as well as the configuration itself, is optional, and the only place where you can make a mistake and get an error is the dependency name.

packages/host/src/node/path-utils.test.ts
owl352: if the number is even, then -1, if it is not even, then +1

packages/host/src/node/path-utils.ts
kraenhansen: I mean, someone could put some non-object value in the `reactNativeNodeApi` field or a non-array in the `reactNativeNodeApi.scan.dependencies` or a non-string as an element of `reactNativeNodeApi.scan.dependencies` - all very unlikely, but ideally, that shouldn't result in a catastrophic failure.

packages/host/src/node/path-utils.test.ts
kraenhansen: I'd appreciate a local variable and comment to help us understand the meaning of this a few months from now? 🙂

owl352: done

packages/host/src/node/path-utils.ts
owl352: done ✅

packages/host/src/node/path-utils.ts
owl352: done ✅

PR #372, opened 2026-07-13
Adopt Hermes' first-party Node-API (static_h)

kraenhansen:
Migrates from the Microsoft/`kraenhansen/hermes`-fork Hermes patch flow to Hermes' **first-party Node-API**, merged on `facebook/hermes` branch `static_h` under `API/napi/` (target `hermesNapi`, NAPI v10, built by default). RN has not pulled this in yet, so Hermes is built from source at a pinned `static_h` SHA.

### What works now
- **Real Node-API env on iOS sim.** `CxxNodeApiHostModule` casts the JSI runtime to `IHermes`, reads the low-level `vm::Runtime*` via `getVMRuntimeUnsafe()`, and creates the env with `hermes_napi_create_env(vm, nullptr)`. One env is created per addon, as in Node (`napi_module_register_by_symbol` creates "a new napi_env for this specific module"), and cached on the addon record — the env carries addon-scoped state such as instance data. Each env is owned by the runtime and torn down with it.
- `MOCHA_REMOTE_CONTEXT=allTests` → **14 passing** on iOS sim: all `node-addon-examples` getting-started addons (napi + node-addon-api + Rust `ferric-example`), plus `buffers`, `async`, and a `js-native-api` node-test.

### Linking `hermesNapi` (no pod-side surgery needed)
The RN `hermesvm` framework already `-force_load`s `hermesNapi`, so all `napi_*` runtime symbols are exported and `hermes_napi_create_env`'s code is present. The public `hermes_napi_*` entry points are exported from the framework as long as Hermes is built from a checkout that includes [facebook/hermes#2044](https://github.com/facebook/hermes/pull/2044) *"Export public hermes_napi entry points with NAPI macros"* — which the pinned SHA already contains. No force_load, no second VM copy, no source patching in this repo. (During bring-up a **stale, pre-#2044 vendored checkout** briefly stripped the symbol under `-fvisibility=hidden`; the fix is simply to ensure the vendored checkout is actually at the pinned SHA.)

### Clean break
Drops RN 0.79–0.81 support, the `kraenhansen/hermes` fork tags, JSI-header copying, and `rt.createNodeApiEnv()`. Targets RN `0.87.0-nightly`. Old npm releases keep serving old RN versions.

### Test plan
- [x] `npm run build && npm run prettier:check`
- [x] Workspace unit tests (`react-native-node-api`, `cmake-rn`, `gyp-to-cmake`, `cmake-file-api`)
- [x] `npm run lint`
- [x] iOS sim e2e (`MOCHA_REMOTE_CONTEXT=allTests`) — 14 passing
- [ ] Android emulator e2e
- [ ] macOS


owl352:
> Reference hermes_napi_host (CallInvoker + worker pool) for async work / thread-safe functions

I think I can test this part on real module when it's ready. Asynchronous bindings were added in the previous update of our sdk.

kraenhansen:
I concluded merging this will break MacOS again, until `react-native-macos` ship a release based on React Native ≥ 0.82 (see https://github.com/callstackincubator/react-native-node-api/issues/392 for details)

PR #401, opened 2026-08-12
ci: upgrade to Changesets v3 and changesets/action v2

### LINKED ISSUE #397: Upgrade to Changesets v3 and changesets/action v2
Follow-up to #396, which worked around the release failure rather than fixing its cause.

## Background

`changeset publish` on the 2.x line that was pinned here shells out to `npm info <pkg> --json` to find out what's already on the registry — there was no pnpm code path. npm ≥ 11 validates the root `package.json`'s `devEngines` on *every* command, so `npm info` exited 1 with `EBADDEVENGINES` ("Invalid name `pnpm` does not match `npm`") and the publish died before it started. #396 unblocked releases in the meantime with `npm_config_force: true` on the changesets step.

`@changesets/cli` v3 and `changesets/action` v2 are both stable now (v3.0.0 and v2.0.0 respectively, as of 2026-08-11/12) and are implemented in #401, which does the coordinated bump this issue called for:

- `@changesets/cli` → v3
- `changesets/action@v1` → v2, with the input renames that go with it (`publish:` → `publish-script:`, `GITHUB_TOKEN` env → the `github-token:` input, defaulting to `${{ github.token }}`)
- `.changeset/config.json`'s `$schema` bumped to `@changesets/config@4.0.0`

## `npm_config_force: true` is not removable by this upgrade

This issue originally expected the v3 upgrade to let `npm_config_force: true` be dropped, on the theory that v3's publish pipeline is package-manager-aware and would run `pnpm info`/`pnpm pack`/`pnpm publish` without ever invoking npm. That turned out to be half right: **`pnpm pack` and `pnpm publish` don't shell out to npm** (verified with `strace`), but **`pnpm info` does** — `pnpm info <pkg>` directly `execve`s `npm info <pkg>` under the hood, since pnpm implements that subcommand as a passthrough to the real npm CLI rather than its own registry client. `changeset publish`/`changeset publish-plan` still call `info` to check what's already published, so the same `EBADDEVENGINES` failure #396 worked around still fires — with v3, exactly as with v2.x.

So `npm_config_force: true` needs to stay on the changesets step indefinitely; there's no changesets- or pnpm-version bump on the horizon that removes the need for it (short of pnpm reimplementing `info` without shelling to npm, which isn't tracked anywhere). #401 updates the workflow comment accordingly instead of leaving it framed as a removal condition.

## Related

- Workaround (stays): #396
- Trusted publishing / OIDC setup: #399
- Implementation: #401

kraenhansen:
Fixes #397.

## What moved

Both packages were blocked on prereleases when #397 was filed; both have since shipped stable, so this does the coordinated bump it called for:

- `@changesets/cli` → `^3.0.0` (root `devDependencies`)
- `changesets/action@v1` → `v2` in `.github/workflows/release.yml`
- Renamed the step's inputs to v2's kebab-case scheme: `publish:` → `publish-script:`. Dropped the `GITHUB_TOKEN` env var — v2 reads it through a `github-token:` input that already defaults to `${{ github.token }}`, same value.
- `.changeset/config.json`: bumped the `$schema` pin to `@changesets/config@4.0.0` (what `@changesets/cli@3.0.0` depends on). None of v4's breaking changes touch our config — no `prettier` key, `baseBranch` is already `"main"`, no experimental snapshot option in use, and `changeset tag` (renamed to `changeset git-tag`) isn't used anywhere in our scripts.

Both had to move together, as #397 flagged: v3 dropped the `New tag: <pkg>@<version>` stdout line `changesets/action@v1` parses to detect what published; the action's v2 line instead reads NDJSON `git-tag` events the CLI writes to `$CHANGESETS_OUTPUT`. CLI v3 with action v1 would publish but report `published: false` and push no tags/releases — a silent regression.

## Correcting an assumption in #397

The issue expected this upgrade to let `npm_config_force: true` (added in #396) be removed, reasoning that v3's publish pipeline is pnpm-aware and would never invoke npm. That's true for `pnpm pack`/`pnpm publish` — verified here with `strace`, neither execs `npm`. But `pnpm info` (what `changeset publish`/`publish-plan` use to check what's already on the registry) is itself a passthrough to the real npm CLI: `strace` shows `pnpm info <pkg>` directly `execve`s `npm info <pkg>`. That still trips npm ≥ 11's `devEngines` validation of this repo's own root `package.json` (`devEngines.packageManager: pnpm`, but the process running is npm), the exact failure #396 worked around.

So `npm_config_force: true` stays — there's no changesets- or pnpm-version bump that removes the need for it. I updated the workflow comment to reflect this instead of leaving it as a removal condition that isn't actually reachable. I also updated #397 itself so it reflects this rather than the original "prereleases are the only blocker" framing.

## Verification

On Node 24 / pnpm 10.33 (this repo's pinned versions):
- `pnpm exec changeset status` and `pnpm exec changeset publish-plan` — the latter reproduces `EBADDEVENGINES` without `npm_config_force`, and completes cleanly with it (same shape as the pre-upgrade behavior), confirming the escape hatch is still both necessary and sufficient.
- `pnpm run build`, `pnpm run prettier:check` — pass.
- `pnpm run lint` — pre-existing, unrelated failures in `apps/test-app/App.tsx` (confirmed present on `main` without this change).
- `pnpm test` — pre-existing, unrelated failures in `packages/host`'s permission-based path-utils tests (they assume file permissions restrict access, which doesn't hold running as root in this container; confirmed present on `main` without this change).

**Not verified**: the actual GitHub Actions publish path. The input/output renames and `github-token` default only get exercised once a "Version Packages" PR merges and the Release job runs the publish step for real, so this needs a live run to confirm end-to-end.

---
_Generated by [Claude Code](https://claude.ai/code/session_015L6NVjP52doXVWZzNyx5L3)_

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://unpkg.com/@changesets/config@3.1.1/schema.json",
+  "$schema": "https://unpkg.com/@changesets/config@4.0.0/schema.json",
   "changelog": "@changesets/cli/changelog",
   "commit": false,
   "fixed": [],
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -58,19 +58,19 @@ jobs:
 
       - name: Create Release Pull Request or Publish to NPM
         id: changesets
-        uses: changesets/action@v1
+        uses: changesets/action@v2
         with:
-          publish: pnpm run release
+          publish-script: pnpm run release
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          # Changesets shells out to `npm info` (there is no pnpm equivalent it
-          # can use) to determine what is already published. npm ≥ 11 checks the
-          # root package.json's "devEngines" on every command and errors with
-          # EBADDEVENGINES because we require pnpm as the package manager, which
-          # fails the publish before it starts. Passing force downgrades that
-          # check to a warning; the actual publishing still runs through
-          # `pnpm publish`. Changesets v3 fixes this properly by running
-          # `pnpm info` when pnpm is detected, but it needs changesets/action v2
-          # to go with it and both are prereleases for now — remove this once we
-          # upgrade to those.
+          # changeset publish detects the pnpm lockfile and correctly shells out
+          # to `pnpm info`/`pnpm pack`/`pnpm publish` instead of npm's — but
+          # `pnpm info` is itself a documented passthrough to the real npm CLI
+          # (confirmed by strace: `pnpm info <pkg>` execs `npm info <pkg>`), so
+          # it still runs into npm ≥ 11 validating this repo's own root
+          # package.json on every invocation and failing with EBADDEVENGINES
+          # (we declare devEngines.packageManager: pnpm, but the process
+          # running is npm). `pnpm pack`/`pnpm publish` don't shell out to npm
+          # and are unaffected. Passing force downgrades the devEngines error
+          # to a warning; this is pnpm's own `info` implementation, so there's
+          # no changesets- or pnpm-version bump that removes the need for it.
           npm_config_force: true
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
   ],
   "license": "MIT",
   "devDependencies": {
-    "@changesets/cli": "^2.29.5",
+    "@changesets/cli": "^3.0.0",
     "@eslint/js": "^9.32.0",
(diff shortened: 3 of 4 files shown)


kraenhansen:
@kraenhansen for review — couldn't request you formally since the PR is authored as you.

One thing worth double-checking on your side: I couldn't exercise the actual publish step through `changesets/action` itself (only the underlying `changeset publish-plan`/CLI calls locally), so the `github-token` default and the `publish-script` rename are unverified against a real Action run until the next "Version Packages" PR merges.

---

Issue #414, opened 2026-08-12
Run linting without building the native code


kraenhansen:
Tracking two related `TODO`s in the `lint` job of `.github/workflows/check.yml`.

The lint job currently has to set up a full native toolchain — JDK 17, the Android SDK + NDK, and the `x86_64-linux-android` Rust target — and then bootstrap two packages, purely so that type-checking has types to work with:

https://github.com/callstackincubator/react-native-node-api/blob/29a527d87fa8cb965b0048b1b4c1cff2b0eab281/.github/workflows/check.yml#L62-L63

https://github.com/callstackincubator/react-native-node-api/blob/29a527d87fa8cb965b0048b1b4c1cff2b0eab281/.github/workflows/check.yml#L76-L77

That is a lot of wall-clock time and cache surface for a job whose actual output is ESLint, Prettier, depcheck and publint results. It also couples the fastest-feedback job to the health of the native toolchain: an NDK or `rustup` hiccup fails linting on a pure-TypeScript PR.

Two things need solving, and they can be tackled independently:

1. **`weak-node-api`** — `pnpm --filter weak-node-api run bootstrap` is run for its generated types. Splitting the generation of the TypeScript/C++ declarations from the actual native build would let the lint job generate types only, dropping the JDK/Android SDK/NDK setup.
2. **`ferric-example`** — as the existing `TODO` suggests, either add an option to `ferric` to emit only the `.d.ts` (skipping the Cargo build), or commit the generated types into the repo as a fixture — the latter doubles as the fixture an `init` command would need (#299).

Once both are addressed, the JDK, Android SDK and `rustup target add` steps can be removed from the lint job.

kraenhansen:
Resolved by #435, merged to `next`:

- `weak-node-api`: the lint job now runs `prebuild:prepare` (header copy + C++/TS declaration codegen) instead of `bootstrap`, dropping the native CMake build.
- `ferric-example`: went with the "commit the generated types as a fixture" option rather than a types-only build — `ferric build --dts-only` was added too (for regenerating the fixture by hand), but napi-rs's typegen turned out to always run a real `cargo build` under the hood, so a checked-in fixture is what actually keeps the lint job off the Rust toolchain. Two label-gated jobs that do a real `ferric build` now `git diff --exit-code` the fixture to catch drift.

The JDK 17, Android SDK, and `rustup target add` steps are gone from `.github/workflows/check.yml`'s `lint` job on `next`. Closing.

---
[...]

PR #435, opened 2026-08-13
ci: run linting without building native code

kraenhansen:
## What

Resolves both `TODO`s in the `lint` job of `.github/workflows/check.yml`, so the job no longer needs a native (JDK/Android SDK/NDK/Rust-Android) toolchain just to type-check TypeScript.

1. **`weak-node-api`** — the lint job now runs the existing `pnpm --filter weak-node-api run prebuild:prepare` script instead of `bootstrap`. `prebuild:prepare` (header copy + C++/TS declaration codegen) already didn't touch anything native — it only needs `clang-format`, which the job already installs via `aminya/setup-cpp`. `bootstrap` additionally ran `prebuild:build` (the actual CMake/NDK/Xcode native compile), which was never needed for typing.

2. **`ferric-example`** — added a new `ferric build --dts-only` flag. It runs a plain host `cargo build` (napi-rs's typedef codegen) to emit the crate's `.d.ts` and JS entrypoint, without cross-compiling any Android/Apple binaries. The output library basename is now derived from `cargo metadata`'s cdylib target name instead of from the paths of already-built platform artifacts (previously `determineLibraryBasename` would `assert()`-fail on an empty array whenever `ANDROID_HOME`/darwin weren't present, since `ferric build`'s target auto-detection produced zero targets). Wired up as `ferric-example`'s new `build:types` script.

With both native builds no longer needed, the JDK 17, Android SDK, and `rustup target add x86_64-linux-android` steps are removed from the `lint` job entirely.

## What CI will actually verify

This is a CI-infrastructure change — the important verification (that the new lint job steps work correctly on GitHub's runners *without* the JDK/Android SDK/NDK installed) can only be confirmed by this PR's own `lint` job run, not by me locally. What I *did* verify locally on a Linux worker with Node 24 and `cargo` present, but **no Android SDK, NDK, or JDK**, reproducing the new job's steps end-to-end:

```
pnpm install
pnpm run build
pnpm --filter weak-node-api run prebuild:prepare
pnpm --filter @react-native-node-api/ferric-example run build:types
pnpm run lint
pnpm run prettier:check
pnpm run depcheck
pnpm run publint
```

All of these passed, and the generated `ferric_example.d.ts`/`.js` matched what a full `ferric build` would produce (`export declare function sum(a: number, b: number): number`). I also confirmed the previous behavior actually failed the way I expected: `ferric build` with no explicit targets on this same toolchain-less box throws `AssertionError: Expected at least one library path to determine its basename` (from `determineLibraryBasename`), which is exactly the bug the JDK/Android SDK setup was masking by making `ANDROID_HOME` available.

I could not exercise the actual GitHub Actions runner environment, self-hosted caching, or the interaction with `ccache`/`aminya/setup-cpp` in situ — that's what this PR's CI run itself needs to confirm.

Addresses #414. Confident enough in both halves that this should let the whole issue close once CI is green — flagging as "addresses" rather than "closes" until the actual CI run confirms it holds up on GitHub's infrastructure.

---

---


kraenhansen:
I suspect we cannot trust that `generateTypeScriptDeclarations` doesn't actually drive a build of native code.

kraenhansen:
Confirmed the suspicion — `napi-rs`'s `napi build` has no typegen-only mode. Running `--dts-only` locally leaves a fully populated ~123MB `target/` directory behind, including a compiled `libferric_example.so`, so it's a real host `cargo build`, not a no-op. "Skipping the native build entirely" in the flag's original description was wrong; only Android/Apple cross-compilation was actually skipped, so the lint job stayed coupled to the host Rust toolchain's health — exactly what #414 was trying to get away from.

Pushed a fix (63f12e8) that switches to the issue's other suggested option: `ferric_example.d.ts`/`.js` are now committed as a fixture (no longer gitignored), and the lint job no longer runs any `ferric` command at all. `--dts-only` stays as the way to regenerate the fixture by hand, with its description corrected to say what it actually does.

To guard against the fixture drifting from the Rust source, the two CI jobs that already do a real `ferric build` (Android and Apple triplets) now `git diff --exit-code` the two committed files right after building. Caveat: both of those jobs are label-gated (`Android 🤖` / `Ferric 🦀`) rather than running on every PR, so this doesn't fully close the gap for an unlabeled PR that changes `packages/ferric-example/src/lib.rs` — flagging that rather than papering over it.

---

---

PR #436, opened 2026-08-13
docs: instruct Claude Code to attach CI labels when opening PRs

kraenhansen:
## Summary

`.github/workflows/check.yml`'s `pull_request` trigger fires only on `opened`, `synchronize` and `reopened` — not `labeled`. Several jobs are gated on a label being present in that triggering event's payload (`host`, `Apple 🍎`, `Android 🤖`, `MacOS 💻`, `Ferric 🦀`, `weak-node-api`). Since `create_pull_request` (the GitHub MCP tool used to open PRs) has no `labels` parameter, labels can only be attached via a follow-up call after the PR already exists — which means the gated jobs silently never run against that PR unless something pushes a new commit afterward.

This surfaced concretely on #434: it touches `packages/host/cpp/RuntimeNodeApi.{cpp,hpp}` and needed `host`/`Apple 🍎`/`Android 🤖` for real coverage, but the labels were only attached after the PR was opened, so the Check run had already evaluated those `if:` conditions without them.

Adds `.claude/CLAUDE.md` documenting:
- why a label added after PR creation doesn't retroactively trigger gated jobs
- a label → job → "attach when the diff touches…" table for the six label-gated jobs
- the required follow-up: attach labels, then push one more commit (even a trivial/empty one) so a `synchronize` event re-evaluates the gated jobs with labels present

This is a base against `main` per the request that prompted it — not `next`, since the repo's other content conventions live there but `.claude/` is tooling/agent-instruction surface rather than package code.

## Testing

- `pnpm exec prettier --check .claude/CLAUDE.md` — passes (also reformatted the table via `--write` before committing).

---

---

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -1,0 +1,55 @@
+# CLAUDE.md (`.claude/`)
+
+Guidance specific to Claude Code sessions that open pull requests against this
+repository (including automated/scheduled sessions). See the root `CLAUDE.md`
+and `AGENTS.md` for everything else.
+
+## Attach CI labels when you open a PR
+
+`.github/workflows/check.yml`'s `pull_request` trigger only fires on
+`opened`, `synchronize` and `reopened` — **not** `labeled`. Several of its
+jobs are gated behind a label check evaluated against that triggering
+event's payload, e.g.:
+
+```yaml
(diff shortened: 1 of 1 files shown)


kraenhansen:
Corrected the label table: it's `Host 🏡`, not `host` — the plain-`host` label just got deleted from the repo. That mismatch was also a real bug in `.github/workflows/check.yml`'s `host-cpp-tests` condition, fixed separately on #434.

---

---

PR #438, opened 2026-08-13
Upgrade bufout to v1.0.0 and remove EventEmitter listener limits

kraenhansen:
## Summary

Upgrades `bufout` to v1.0.0, which implements efficient listener management for concurrent child process spawning. This eliminates the need to raise `EventEmitter.defaultMaxListeners` to 100 across the CLI packages.

## Changes

- **Dependency upgrade**: Updated `bufout` from ^0.3.2 to ^1.0.0 in `@react-native-node-api/cli-utils`
- **Removed EventEmitter configuration**: Deleted `EventEmitter.defaultMaxListeners = 100` assignments from:
  - `packages/ferric/src/run.ts`
  - `packages/cmake-rn/src/cli.ts`
  - `packages/host/src/node/cli/program.ts`
- **Removed unused imports**: Cleaned up now-unnecessary `EventEmitter` imports from the above files

## Implementation Details

The new version of `bufout` uses a shared pass-through stream per destination and attaches a single shared `exit`/`SIGINT` listener only while children are running. This keeps the number of listeners constant regardless of concurrent child process count, restoring Node's default listener limit and re-enabling the leak detection warning it provides.

https://claude.ai/code/session_01BAjP89a9VA9EtQsVBxGcto

### DIFF
--- a/.changeset/bufout-one-point-oh.md
+++ b/.changeset/bufout-one-point-oh.md
@@ -1,0 +1,16 @@
+---
+"@react-native-node-api/cli-utils": patch
+"react-native-node-api": patch
+"cmake-rn": patch
+"ferric-cli": patch
+---
+
+Upgrade `bufout` to v1.0.0, which keeps the number of listeners on the process
+and the output streams constant regardless of how many children are spawned
+concurrently: a single shared `exit`/`SIGINT` listener is attached only while
+children are running, and every child pipes into one shared pass-through per
+destination stream.
+
+That removes the reason for the CLIs to raise `EventEmitter.defaultMaxListeners`
+to 100, so those assignments are gone and Node's default limit again applies —
+restoring the leak warning it exists to give.
--- a/packages/cli-utils/package.json
+++ b/packages/cli-utils/package.json
@@ -11,7 +11,7 @@
   },
   "dependencies": {
     "@commander-js/extra-typings": "^14.0.0",
-    "bufout": "^0.3.2",
+    "bufout": "^1.0.0",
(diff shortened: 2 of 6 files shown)


kraenhansen:
Merged `next` in to pick up #432 — the iOS run on the previous head failed for a reason unrelated to this diff.

**What failed:** `Test app (iOS)` built fine and got through `buffers` and `async`, then the app hard-crashed right after loading `threadsafe-function-test--addon` (`DISCONNECTION for no particular reason (code = 1006)`).

**Why it isn't this PR:** `bufout` is host-side build tooling — it spawns `cmake`/`xcodebuild` on the runner and never ends up in the app binary, so it can't affect simulator runtime behaviour. The build itself succeeded and every addon loaded.

**What it actually is:** this branch was based on c73d30c, four commits behind `next`. #432 (`715a24e`), merged since, describes precisely this symptom:

> node-addon-api's tsfn error path (which calls `napi_fatal_exception` whenever an exception escapes a thread-safe-function callback) [had] no chance of being observed or handled — a single throwing tsfn callback hard-killed the app

That PR notes it wasn't exercised on-device, so this run is likely the first device coverage it gets. Treating the merge as the fix is a hypothesis, not a confirmed one — if iOS stays red on the new head, the failure is pre-existing on `next` and belongs in its own issue rather than in a dependency bump.

Labels `Apple 🍎` / `Android 🤖` / `Ferric 🦀` are attached deliberately: those jobs are the only place the changed spawn machinery runs real concurrent builds.

---

PR #448, opened 2026-08-15
Close two test gaps: unverified prebuild Info.plist and a missing .node fixture

### LINKED ISSUE #424: Two test gaps: unverified prebuild `Info.plist` and a missing `.node` fixture in the Babel plugin tests

kraenhansen:
Closes #424.

## `verify-prebuilds` now reads the `Info.plist` it finds

`verifyApplePrebuild` used to `continue` past `Info.plist` without reading it. It now parses it with `@expo/plist` (already a dependency elsewhere in the repo) and asserts:

- `CFBundleExecutable` matches the framework's library name (the `.framework` directory's basename)
- `CFBundleIdentifier` matches `escapeBundleIdentifier("com.callstackincubator.node-api.<libraryName>")` — the default `writeFrameworkInfoPlist` (`packages/host/src/node/prebuilds/apple.ts`) writes when no `--apple-bundle-identifier` is passed, which is how `node-addon-examples` builds

`escapeBundleIdentifier` is now exported from the package's public `node` entrypoint so the verify script (a `react-native-node-api` consumer, same as any addon author) can import it rather than re-deriving the escaping rule.

## The Babel plugin test now proves its own claim — and that found a real bug

"and does not touch required JS files" asserted the plugin leaves `require('./my-addon')` alone, but the fixture had no `my-addon.*.node` for the plugin to have found in the first place — the assertion passed for the wrong reason. Adding a sibling `my-addon.apple.node` (as the `TODO` asked) makes the test exercise real precedence between a `.js` file and a same-named addon, and it failed: `isNodeApiModule` didn't check whether a same-named `.js`/`.cjs`/`.mjs`/`.json` file would already satisfy `require()` before considering `.node` prebuilds, so the plugin rewrote a call that Node's own resolution would never route to the addon.

Fixed `isNodeApiModule` to defer to a colliding source file when the module path has no explicit `.node` extension, matching Node's own resolution order (`.js`/`.json` before `.node`). Explicit `require('./my-addon.node')` calls are untouched by this check, since there's no resolution ambiguity to defer to.

## Verification

- `pnpm run build`, `pnpm run lint`, `pnpm exec prettier --check` clean
- `packages/host`'s test suite: the `plugin` and `escapeBundleIdentifier` suites pass in full; the only failures are 4 pre-existing permission-based tests that fail in this sandbox because it runs as root (verified they fail identically on `next` before this change)
- Manually exercised `verifyFrameworkInfoPlist` against a real `Info.plist` written by `createAppleFramework` (`install_name_tool` isn't available on Linux, so I read the file straight after it's written) — both the pass and the mismatch-detection path behave as expected
- No changeset for `@react-native-node-api/node-addon-examples`: it's private

cc @kraenhansen for review.

### DIFF
--- a/.changeset/babel-plugin-js-precedence.md
+++ b/.changeset/babel-plugin-js-precedence.md
@@ -1,0 +1,14 @@
+---
+"react-native-node-api": patch
+---
+
+Fix the Babel plugin rewriting `require(...)` calls that resolve to a
+same-named `.js`/`.cjs`/`.mjs`/`.json` file sitting next to a Node-API
+prebuild. Node's own module resolution always picks the source file over a
+`.node` addon in that case, so the plugin now leaves those calls alone
+instead of rewriting them to `requireNodeAddon(...)`, which would have loaded
+the wrong module at runtime.
+
+Also exports `escapeBundleIdentifier`, used internally to derive a
+framework's `CFBundleIdentifier`, so it can be reused to verify one against
+its expected value.
--- a/packages/host/src/node/babel-plugin/plugin.test.ts
+++ b/packages/host/src/node/babel-plugin/plugin.test.ts
@@ -129,7 +129,8 @@ describe("plugin", () => {
     itTransforms("and does not touch required JS files", {
       files: {
         "package.json": `{ "name": "my-package" }`,
-        // TODO: Add a ./my-addon.node to make this test complete
+        "my-addon.apple.node/my-addon.node":
+          "// This is supposed to be a binary file",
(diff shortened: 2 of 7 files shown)


kraenhansen:
Closing this in favor of #450, which builds on @wanxiankai's earlier #426 (the first PR to fix #424) plus the same improvements this PR made (moving the fix into `isNodeApiModule`, zod validation, reusing `escapeBundleIdentifier`). Sorry for the duplicated effort — #426 wasn't spotted before this PR was opened.

---

PR #450, opened 2026-08-16
Fix .js/.node precedence in isNodeApiModule; dedupe Info.plist checks (builds on #426)

### LINKED ISSUE #424: Two test gaps: unverified prebuild `Info.plist` and a missing `.node` fixture in the Babel plugin tests

kraenhansen:
## Summary

This builds directly on @wanxiankai's #426 (branch `fix/verify-prebuild-fixtures`, included here as the first commit) with a few refinements on top. It closes #424, superseding #448 and #449 (two independent attempts at the same fix, opened before this one was aware #426 already existed).

> Note: this PR targets `next` in this repo rather than #426's branch directly, because this session's GitHub access is scoped to `callstackincubator/react-native-node-api` only and can't open a PR against a fork. @kraenhansen — happy to have this merged as-is, or closed in favor of #426 if you'd rather land these follow-ups there instead; wanxiankai's original commit is preserved either way.

## What's on top of #426

- **Move the `.js`/`.node` precedence fix into `isNodeApiModule`** (`packages/host/src/node/path-utils.ts`) instead of guarding the Babel plugin's `require()` rewrite with `createRequire(this.filename).resolve(id)`. Reasons:
  - `isNodeApiModule` is the shared utility (also used by `findNodeAddonForBindings`), so fixing it there benefits every caller instead of only the plugin's `require()` handling.
  - `createRequire(...).resolve()` runs Node's own CommonJS resolution algorithm at Babel-transform time to predict what React Native's Metro bundler will resolve at runtime — but Metro doesn't use Node's resolver (it has its own, with platform extensions, Haste, etc.), so the two can diverge. Checking for a same-named `.js`/`.cjs`/`.mjs`/`.json` sibling file directly is simpler and doesn't depend on that assumption holding.
- **Validate the parsed `Info.plist` with a `zod` schema** instead of ad hoc `"key" in obj` checks on an `unknown` value, matching how the rest of the codebase validates untrusted structured data (`zod` is already a dependency across several packages).
- **Reuse the exported `escapeBundleIdentifier`** (now re-exported from `react-native-node-api`'s `node` entrypoint) instead of re-deriving the bundle-identifier escaping regex inline in `verify-prebuilds.mts`, so the verifier can't silently drift out of sync with what `writeFrameworkInfoPlist` actually writes.

## Test plan

- `pnpm install && pnpm run build` — clean
- `pnpm --filter react-native-node-api test` — the `plugin`, `isNodeApiModule`, and `escapeBundleIdentifier` suites pass; the only failures are 4 pre-existing permission-based tests that fail identically with and without this diff (confirmed via `git stash`) because this sandbox runs as root
- `pnpm exec tsc --build packages/node-addon-examples/tsconfig.node-scripts.json` — clean (type-checks `verify-prebuilds.mts`)
- `pnpm exec eslint` on all changed files — clean
- `pnpm run prettier:check` — clean

Closes #424

cc @kraenhansen for review.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q1k6UQJPPaqKEKmnsRUatt)_

### DIFF
--- a/.changeset/calm-bears-resolve.md
+++ b/.changeset/calm-bears-resolve.md
@@ -2,4 +2,4 @@
 "react-native-node-api": patch
 ---
 
-Preserve Node.js module resolution precedence when a JavaScript file and native addon share a basename.
+Preserve Node.js module resolution precedence when a JavaScript file and native addon share a basename: `require('./foo')` no longer gets rewritten to load a Node-API addon when a same-named `foo.js`/`.cjs`/`.mjs`/`.json` file exists alongside it, since that source file is what `require()` actually resolves to. An explicit `require('./foo.node')` is unaffected.
--- a/.changeset/calm-bears-resolve.md
+++ b/.changeset/calm-bears-resolve.md
@@ -1,0 +1,5 @@
+---
+"react-native-node-api": patch
+---
+
+Preserve Node.js module resolution precedence when a JavaScript file and native addon share a basename.
--- a/packages/host/src/node/babel-plugin/plugin.test.ts
+++ b/packages/host/src/node/babel-plugin/plugin.test.ts
@@ -129,7 +129,7 @@ describe("plugin", () => {
     itTransforms("and does not touch required JS files", {
       files: {
         "package.json": `{ "name": "my-package" }`,
-        // TODO: Add a ./my-addon.node to make this test complete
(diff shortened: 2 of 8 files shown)


kraenhansen:
Closing as I pushed to https://github.com/callstackincubator/react-native-node-api/pull/426 instead.

